PR #3, opened 2022-09-14
tests: Wait for test process to be started

bdrung:
The test case `test_omit_all_processes_except_one` is racy and fails sometimes:

```
test_omit_all_processes_except_one (tests.system.test_unkillable_shutdown.TestUnkillableShutdown)
unkillable_shutdown will write exactly one report. ... Running as unit: run-r0b0d3fe87473410c9eb61a5ae4635938.service
FAIL
======================================================================
FAIL: test_omit_all_processes_except_one (tests.system.test_unkillable_shutdown.TestUnkillableShutdown)
unkillable_shutdown will write exactly one report.
----------------------------------------------------------------------
  File "tests/system/test_unkillable_shutdown.py", line 96, in test_omit_all_processes_except_one
    pid = self._launch_process_with_different_session_id(existing_pids)
  File "tests/system/test_unkillable_shutdown.py", line 90, in _launch_process_with_different_session_id
    self.assertEqual(len(test_executable_pids), 1, test_executable_pids)
AssertionError: 0 != 1 : set()

----------------------------------------------------------------------
```

Wait for test process to be started.

Bug: https://launchpad.net/bugs/1989365

### DIFF
--- a/tests/system/test_unkillable_shutdown.py
+++ b/tests/system/test_unkillable_shutdown.py
@@ -66,14 +66,13 @@ class TestUnkillableShutdown(unittest.TestCase):
     @unittest.skipIf(
         get_init_system() != "systemd", "running init system is not systemd"
     )
-    def _launch_process_with_different_session_id(
-        self, existing_pids: list
-    ) -> int:
+    def _launch_process_with_different_session_id(self) -> int:
         """Launch test executable with different session ID.
 
         getsid() will return a different ID than the current process.
         """
         service_manager = "--system" if os.geteuid() == 0 else "--user"
+        existing_pids = self._get_all_pids()
         try:
             subprocess.run(
                 ["systemd-run", service_manager, self.TEST_EXECUTABLE]
@@ -93,7 +92,7 @@ class TestUnkillableShutdown(unittest.TestCase):
     def test_omit_all_processes_except_one(self):
         """unkillable_shutdown will write exactly one report."""
         existing_pids = self._get_all_pids()
-        pid = self._launch_process_with_different_session_id(existing_pids)
+        pid = self._launch_process_with_different_session_id()
         try:
             self._call(omit=existing_pids, expected_stderr="")
         finally:
@@ -106,8 +105,7 @@ class TestUnkillableShutdown(unittest.TestCase):
     def test_write_reports(self):
         """unkillable_shutdown will write reports."""
         # Ensure that at least one process is honoured by unkillable_shutdown.
-        existing_pids = self._get_all_pids()
-        pid = self._launch_process_with_different_session_id(existing_pids)
+        pid = self._launch_process_with_different_session_id()
         try:
             self._call()
         finally:
--- a/tests/system/test_unkillable_shutdown.py
+++ b/tests/system/test_unkillable_shutdown.py
@@ -14,6 +14,7 @@ import shutil
 import signal
 import subprocess
 import tempfile
+import time
 import typing
 import unittest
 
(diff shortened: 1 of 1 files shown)


tests/system/test_unkillable_shutdown.py
```diff
@@ -82,13 +83,25 @@ def _launch_process_with_different_session_id(self) -> int:
...
+
+    def _wait_for_process(self, program: str, timeout_sec=5.0) -> int:
+        """Wait until one process with the given name is running."""
+        timeout = 0.0
+        while timeout < timeout_sec:
+            pids = {
+                pid for pid in pidof(program) if pid not in self.existing_pids
+            }
+            if pids:
+                self.assertEqual(len(pids), 1, pids)
```
schopin-pro, line 96: suggestion (non-blocking): check the args of the individual processes to find `self.TEST_ARGS`

Having random `sleep` popping up on a system is not really uncommon, so rather than hoping that we're the only ones using that binary, we could check that `/proc/$PID/cmdline` actually matches our own specific duration.

tests/system/test_unkillable_shutdown.py
```diff
@@ -82,13 +83,25 @@ def _launch_process_with_different_session_id(self) -> int:
...
         except FileNotFoundError as error:  # pragma: no cover
             self.skipTest(f"{error.filename} not available")
-        test_executable_pids = {
-            pid
-            for pid in pidof(self.TEST_EXECUTABLE)
-            if pid not in self.existing_pids
-        }
-        self.assertEqual(len(test_executable_pids), 1, test_executable_pids)
-        return test_executable_pids.pop()
+        return self._wait_for_process(self.TEST_EXECUTABLE)
```
schopin-pro, line 86: question (non-blocking): isn't systemd-run already supposed to only exit once its child process runs?

schopin-pro (approved):
LGTM, as it's a strict improvement on the current state of affairs.

I still find the approach fragile, see my suggestion below, but that's non-blocking.

All inline comments use [Conventional Comments](https://conventionalcomments.org/] prefixes.

tests/system/test_unkillable_shutdown.py
bdrung: The man-page of systemd-run says:
> the start request for the transient unit will be verified, enqueued and systemd-run will wait until the unit's start-up is completed.

That sounds like systemd-run should wait long enough, but the autopkgtest failure proves the opposite.

tests/system/test_unkillable_shutdown.py
bdrung: Yes, that a good idea. Idea for implementation: Write a pidof function using the psutil library that takes the full arguments.

tests/system/test_unkillable_shutdown.py
```diff
@@ -85,17 +85,16 @@ def _launch_process_with_different_session_id(
...
         return test_executable_pids.pop()
 
     def test_omit_all_processes_except_one(self):
         """unkillable_shutdown will write exactly one report."""
-        existing_pids = self._get_all_pids()
-        pid = self._launch_process_with_different_session_id(existing_pids)
+        pid = self._launch_process_with_different_session_id()
         try:
-            self._call(omit=existing_pids, expected_stderr="")
+            self._call(omit=self.existing_pids, expected_stderr="")
```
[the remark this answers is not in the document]
bdrung: The test case will be executed after the setUp. So the time difference will be negligible, but I will change the MR to make that easier.

schopin-pro (approved):
Good to merge, the psutils-based solution can be done later if this still gives us grief.

PR #6, opened 2022-09-15
Disable debuginfod when collecting a report

sergiodj:
With the new Ubuntu debuginfod
service (https://debuginfod.ubuntu.com), and with the prospect of
having the system automatically fetch debuginfo from the internet
without user intervention, it is necessary to adjust apport to cope
with this scenario.

I had a conversation with bdmurray and he was concerned that having a
debuginfod-enabled GDB generate the corefiles that are eventually
submitted to Ubuntu can be a problem.

This commit disables GDB's debuginfod support when collecting a
report.  I chose not to touch GDB's settings on apport-retrace because
I think we should honour the user choice there.

Bug: https://launchpad.net/bugs/1989803

### DIFF
--- a/apport/report.py
+++ b/apport/report.py
@@ -887,8 +887,14 @@ class Report(problem_report.ProblemReport):
                 "gdb not found in retracing env",
             )
 
-        # limit maximum backtrace depth (to avoid looped stacks)
-        gdb_cmd += ["--batch", "--ex", "set backtrace limit 2000"]
+        gdb_cmd += [
+            "--batch",
+            # limit maximum backtrace depth (to avoid looped stacks)
+            "--ex",
+            "set backtrace limit 2000",
+            "-iex",
+            "set debuginfod enable off",
+        ]
 
         value_keys = []
         # append the actual commands and something that acts as a separator
--- a/bin/apport-valgrind
+++ b/bin/apport-valgrind
@@ -203,8 +203,10 @@ except (KeyboardInterrupt, SystemExit):
     sys.stderr.write("\nInterrupted while preparing to create sandbox\n")
     _exit_on_interrupt()
 
+valgrind_env = {k: v for k, v in os.environ.items() if k != "DEBUGINFOD_URLS"}
+
 try:
-    subprocess.call(argv)
+    subprocess.call(argv, env=valgrind_env)
 except (KeyboardInterrupt, SystemExit):
     sys.stderr.write("\nInterrupted while running valgrind\n")
     _exit_on_interrupt()
--- a/tests/integration/test_apport_valgrind.py
+++ b/tests/integration/test_apport_valgrind.py
@@ -21,6 +21,7 @@ class T(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         cls.env = os.environ | local_test_environment()
+        cls.env.pop("DEBUGINFOD_URLS", None)
 
     def setUp(self):
         self.workdir = tempfile.mkdtemp()
--- a/tests/integration/test_report.py
+++ b/tests/integration/test_report.py
@@ -662,6 +662,8 @@ int main() { return f(42); }
                 [
                     "gdb",
                     "--batch",
+                    "-iex",
+                    "set debuginfod enable off",
                     "--ex",
                     f"run{' ' + ' '.join(args) if args else ''}",
                     "--ex",
@@ -909,6 +911,8 @@ int main() { return f(42); }
                     [
                         "gdb",
                         "--batch",
+                        "-iex",
+                        "set debuginfod enable off",
                         "--ex",
                         f"run {script}",
                         "--ex",
--- a/tests/integration/test_signal_crashes.py
+++ b/tests/integration/test_signal_crashes.py
@@ -874,7 +874,16 @@ class T(unittest.TestCase):
         st = os.stat(core_path)
         self.assertGreater(st.st_size, 10000)
         gdb = subprocess.run(
-            ["gdb", "--batch", "--ex", "bt", command, core_path],
+            [
+                "gdb",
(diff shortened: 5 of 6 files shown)


bdrung (changes requested):
The change looks good. Please also update the gdb calls in the integration tests:
* tests/integration/test_report.py
* tests/integration/test_signal_crashes.py
* tests/integration/test_ui.py

The integration tests must not talk to the Internet.

sergiodj:
> The change looks good. Please also update the gdb calls in the integration tests:
> 
>     * tests/integration/test_report.py
> 
>     * tests/integration/test_signal_crashes.py
> 
>     * tests/integration/test_ui.py
> 
> 
> The integration tests must not talk to the Internet.

Thanks for the review.  I did the changes above, and also expanded the commit a little bit.  valgrind also supports debuginfod, so I went ahead and disabled the feature on it as well.  Unfortunately, valgrind doesn't provide an easy flag to disable debuginfod usage, so I had to resort to unsetting the `DEBUGINFOD_URLS` environment variable.

sergiodj:
@bdrung This is ready to be reviewed again.  Please let me know what you think.  Thanks!

bin/apport-valgrind
```diff
@@ -203,8 +203,10 @@ except (KeyboardInterrupt, SystemExit):
     sys.stderr.write("\nInterrupted while preparing to create sandbox\n")
     _exit_on_interrupt()
 
+newenviron = os.environ.copy() | {"DEBUGINFOD_URLS": ""}
```
bdrung, line 206: This only sets the environment variable to `""` instead of unsetting it. This would be:
```
{k, v: for k, v in os.environ.items() if k != "DEBUGINFOD_URLS"}
```

bdrung, line 206: `newenviron` is not very expressive. The `new` is confusing there. How about `valgrind_env` instead (or just `env`)?

tests/integration/test_apport_valgrind.py
```diff
@@ -40,10 +40,11 @@ def test_valgrind_min_installed(self):
         self.assertIn("tests", out)
 
     def _call(self, argv):
+        newenv = self.env | {"DEBUGINFOD_URLS": ""}
```
bdrung, line 43: Same remarks as above.

apport/report.py
```diff
@@ -890,6 +890,9 @@ def add_gdb_info(self, rootdir=None, gdb_sandbox=None):
         # limit maximum backtrace depth (to avoid looped stacks)
         gdb_cmd += ["--batch", "--ex", "set backtrace limit 2000"]
 
+        # disable debuginfod
+        gdb_cmd += ["-iex", "set debuginfod enable off"]
```
bdrung, line 894: How about combining those two commands into one:
```suggestion
        gdb_cmd += [
            "--batch",
            # limit maximum backtrace depth (to avoid looped stacks)
            "--ex",
            "set backtrace limit 2000",
            "-iex",
            "set debuginfod enable off",
        ]
```

apport/report.py
```diff
@@ -890,6 +890,9 @@ def add_gdb_info(self, rootdir=None, gdb_sandbox=None):
         # limit maximum backtrace depth (to avoid looped stacks)
         gdb_cmd += ["--batch", "--ex", "set backtrace limit 2000"]
 
+        # disable debuginfod
```
bdrung, line 893: The comment "disable debuginfod" does not provide any additional information. This comment can be dropped unless it explains the "why?".

bdrung (changes requested):
You had to touch the gdb call in multiple places. I am not sure if we should refactor apport to only need to touch the gdb arguments in one place.

sergiodj:
Thanks for the review.  I've addressed all the points, hopefully this is good now.

tests/integration/test_apport_valgrind.py
```diff
@@ -40,10 +40,13 @@ def test_valgrind_min_installed(self):
...
 
     def _call(self, argv):
+        valgrind_env = dict(
+            {k, v} for k, v in os.environ.items() if k != "DEBUGINFOD_URLS"
+        )
         process = subprocess.run(
             argv,
             check=False,
-            env=self.env,
+            env=valgrind_env,
```
bdrung, line 50: You can modify `self.env` in `setUpClass` where we already modify the environment variable.

bin/apport-valgrind
```diff
@@ -203,8 +203,12 @@ except (KeyboardInterrupt, SystemExit):
     sys.stderr.write("\nInterrupted while preparing to create sandbox\n")
     _exit_on_interrupt()
 
+valgrind_env = dict(
+    {k, v} for k, v in os.environ.items() if k != "DEBUGINFOD_URLS"
+)
```
bdrung, line 208: You can use a dictionary comprehension instead of a list comprehension:
```suggestion
valgrind_env = {k: v for k, v in os.environ.items() if k != "DEBUGINFOD_URLS"}
```

bdrung (changes requested):
Found two more points.

sergiodj:
After a long battle with black and pylint, this is ready for review.

bdrung (approved):

zeehio:
Hi,

I have not been able to find a way to make apport work with the debuginfod to get the full stacktrace.

In the past I was able to provide a patch a package by myself thanks to the apport stack trace, now the symbol table information is not available neither in the GUI nor with 

```
export DEBUGINFOD_URLS="https://debuginfod.ubuntu.com"
apport-retrace --stdout _usr_bin_marco.1000.crash 
```

How am I supposed to get something instead of "No symbol table info available."?

bdrung:
Yes, you should get a stack trace presented instead.

PR #8, opened 2022-09-29
Address easy to fix pydocstyle complains in tests

bdrung:
`pydocstyle` ignores test files by default. Address the pydocstyle complains in the test files that are easy to address. Leave the more time consuming ones for later.


schopin-pro (approved):
LGTM

PR #9, opened 2022-09-29
tests: Use LRU cache for has_internet()

bdrung:
mypy complains:

```
tests/helper.py:40: error: "Callable[[], Any]" has no attribute "cache"
```

Use a LRU cache for `has_internet()` to make the code shorter and easier to read.

### DIFF
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -1,6 +1,7 @@
 """Helper functions for the test cases."""
 
 import contextlib
+import functools
 import importlib.machinery
 import importlib.util
 import os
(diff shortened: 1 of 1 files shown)


tests/helper.py
```diff
@@ -18,26 +19,22 @@ def get_init_system() -> str:
         return comm.read().rstrip()
 
 
-def has_internet():
+@functools.lru_cache(maxsize=1)
```
schopin-pro, line 22: praise: Neat, I didn't know about that decorator :)

schopin-pro (approved):
LGTM.

In particular, anything that removes useless network traffic is a win in my books ;)

PR #11, opened 2022-09-30
Fix mypy for 'suppress' in apport/ui.py

bdrung:
mypy complains:

```
apport/ui.py:971: error: Incompatible types in assignment (expression has type "bool", variable has type "Union[_SUPPRESS_T, str]")
```

The `help` parameter for `add_argument` takes `Optional[Text]`. Therefore set `suppress` to `None` if the help should not be suppressed.


schopin-pro (approved):
LGTM. Thanks for the detailed commit log, 'cause the diff wasn't really self-explanatory ;)

PR #15, opened 2022-10-05
README.md: Document how to create releases

bdrung:
Document how to create releases.

Is it okay to add it to `README.md` or better put it somewhere else?

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,3 +90,36 @@ to it. Please send patches which are applicable to main as merge requests or
 bug reports, so that
 1. other distributions can benefit from them as well, and
 2. you reduce the code delta to upstream.
+
+Creating releases
+=================
+
+This project uses [semantic versioning](https://semver.org/). To create a
+release, increase the version in [apport/ui.py](apport/ui.py) and document the
+noteworthy changes in [NEWS.md](./NEWS.md). Then commit the changes and get them
+reviewed:
+
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -90,3 +90,36 @@ to it. Please send patches which are applicable to main as merge requests or
...
+Creating releases
+=================
+
+This project uses [semantic versioning](https://semver.org/). To create a
+release, increase the version in [apport/ui.py](apport/ui.py) and document the
+noteworthy changes in [NEWS.md](./NEWS.md). Then commit the changes and get them
+reviewed:
+
+```
+version=$(python3 -c "import apport.ui; print(apport.ui.__version__)")
```
schopin-pro, line 103: question: wouldn't that give you the currently-installed version instead?

I'm always a bit fuzzy on Python import behavior, I usually just force it using PYTHONPATH ;-)

schopin-pro (commented):
I think README.md is fine. You could use a README-devel.md if you prefer.

That's a great initiative regardless. I just have a single simple question :)

README.md
bdrung: That gives the local version since Python looks into the current directory before the installed libraries. IIRC only the libraries shipped by Python itself take precedence before local ones.

bdrung:
Then let's keep it in `README.md`.

PR #16, opened 2022-10-05
CI: Remove system installed Apport

bdrung:
Installing Apport with layout `deb` and prefix `/usr` installs the `etc` directory into the local directory instead of `/`.

Set `--root` to `/` to install all files into the system.

Remove system installed Apport to avoid mixing the Apport version to test and the system installed Apport.


schopin-pro (approved):
LGTM, makes a lot of sense.

PR #17, opened 2022-10-05
tests: Do not cover potential unused retry test code

bdrung:
The KDE test cases wait for dialogs to appear and then interact with them. In case they do not appear after a short time period, more time will be waited and retried. This "waiting additional time" is not triggered on every CI run and therefore the code coverage varies.

Do not cover this retry test code to stabilize the test coverage report.


schopin-pro (approved):
Nice catch!

PR #21, opened 2022-10-07
tests: Test wait_for_gdb_child_process() helper method

bdrung:
The `wait_for_gdb_child_process` helper method can success on the first try. Then the "sleep and retry" part of the function is not used which leads to a reduced code coverage.

Write test cases for the `wait_for_gdb_child_process` helper method to have full code coverage for it.


schopin-pro (approved):
Urgh, mock code always makes my head hurt before I can make sense of what's happening...

LGTM

PR #1, opened 2022-07-08
Adding Github backend

EduardGomezEscandell:
Reports bugs as Github Issues to a given repository. This is done with Ubuntu WSL in mind, but can be easily adapted for other purposes.

### Authentication
Uses Github's Device flow:
https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#device-flow

This means that the user is given this message:

> Input the following code in your browser:
> URL: https://github.com/login/device
> Code: B45A-7491

And this is used to authenticate.

PR #26, opened 2022-11-23
Add GitHub backend for WSL and snaps

nteodosio:
This builds on Edu's work on https://github.com/canonical/apport/pull/1, to add this functionality: If a snap contact: field points to Github, report it there.

The snap report can be tested by installing [this snap](https://github.com/canonical/snap-for-apport-reports/) and then running `apport-bug apport-target`.

### DIFF
--- a/apport/crashdb_impl/github.py
+++ b/apport/crashdb_impl/github.py
@@ -1,0 +1,318 @@
+"""Crash database implementation for Github."""
+
+# Copyright (C) 2022 - 2022 Canonical Ltd.
+# Author: Eduard Gómez Escanell <edu.gomez.escandell@canonical.com>
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation; either version 2 of the License, or (at your
+# option) any later version.  See http://www.gnu.org/copyleft/gpl.html for
+# the full text of the license.
+
+import json
+import time
+from dataclasses import dataclass
+
+import requests
+
+import apport
+import apport.crashdb
+
+
+class Github:
+    """Wrapper around Github API, used to log in and post issues."""
+
+    __last_request: float = time.time()
+
+    def __init__(self, client_id, message_callback):
+        self.__client_id = client_id
+        self.__authentication_data = None
+        self.__access_token = None
+        self.__cooldown = None
+        self.__expiry = None
+        self.message_callback = message_callback
+
+    @staticmethod
+    def _stringify(data: dict) -> str:
+        """Takes a dict and returns it as a string for POSTing."""
+        string = ""
+        for key, value in data.items():
+            string = f"{string}&{key}={value}"
+        return string
+
+    def _post(self, url: str, data: str):
+        """Posts the given data to the given URL.
+        Uses auth token if available"""
+        headers = {"Accept": "application/vnd.github.v3+json"}
+        if self.__access_token:
+            headers["Authorization"] = f"token {self.__access_token}"
+        try:
+            result = requests.post(
+                url, headers=headers, data=data, timeout=5.0
+            )
+        except requests.RequestException as err:
+            self.message_callback(
+                "Failed connection",
+                f"Failed connection to {url}.\n"
+                + "Please check your internet connection and try again.",
+            )
+            raise err
+        finally:
+            self.__last_request = time.time()
+
+        result.raise_for_status()  # Not using UI: the user can't do much here
+        return json.loads(result.text)
+
+    def api_authentication(self, url: str, data: dict):
+        return self._post(url, self._stringify(data))
+
+    def api_open_issue(self, owner: str, repo: str, data: dict):
+        url = f"https://api.github.com/repos/{owner}/{repo}/issues"
+        return self._post(url, json.dumps(data))
+
+    def __enter__(self):
+        """Enters login process. At exit, login process ends."""
+        data = {"client_id": self.__client_id, "scope": "public_repo"}
+        url = "https://github.com/login/device/code"
+        response = self.api_authentication(url, data)
+
+        prompt = (
+            "Posting an issue requires a Github account. If you have "
+            "one, please follow these steps to log in.\n"
+            "\n"
+            "Open the following URL. When requested, write this code "
+            "to enable apport to open an issue.\n"
+            "URL:  {url}\n"
+            "Code: {code}"
+        )
+
+        url = response["verification_uri"]
+        code = response["user_code"]
+
+        self.message_callback(
+            "Login required", prompt.format(url=url, code=code)
+        )
+
+        self.__authentication_data = {
+            "client_id": self.__client_id,
+            "device_code": f'{response["device_code"]}',
+            "grant_type": "urn:ietf:params:oauth:grant-type:device_code",
+        }
+        self.__cooldown = response["interval"]
+        self.__expiry = int(response["expires_in"]) + time.time()
+
+        return self
+
+    def __exit__(self, *_) -> None:
+        self.__authentication_data = None
+        self.__cooldown = 0
+        self.__expiry = 0
+
+    def authentication_complete(self) -> bool:
+        """Asks Github if the user has logged in already.
+        It respects the wait-time requested by Github.
+        """
+        if not self.__authentication_data:
+            raise RuntimeError(
+                "Authentication not started. Use a with statement to do so"
+            )
+
+        t = time.time()
+        waittime = self.__cooldown - (t - self.__last_request)
+        if t + waittime > self.__expiry:
+            self.message_callback(
+                "Failed login",
+                "Github authentication expired. Please try again.",
+            )
+            raise RuntimeError("Github authentication expired")
+        if waittime > 0:
+            time.sleep(waittime)  # Avoids spamming the API
+
+        url = "https://github.com/login/oauth/access_token"
+        response = self.api_authentication(url, self.__authentication_data)
+
+        if "error" in response:
+            if response["error"] == "authorization_pending":
+                return False
+            if response["error"] == "slow_down":
+                self.__cooldown = int(response["interval"])
+                return False
+            raise RuntimeError(f"Unknown error from Github: {response}")
+        if "access_token" in response:
+            self.__access_token = response["access_token"]
+            return True
+        raise RuntimeError(f"Unknown response from Github: {response}")
+
+
+@dataclass(frozen=True)
+class IssueHandle:
+    url: str
+
+
+class CrashDatabase(apport.crashdb.CrashDatabase):
+    """Github crash database.
+    This is a Apport CrashDB implementation for interacting with Github issues
+    """
+
+    def __init__(self, auth_file, options):
+        """Initialize some variables. Login is delayed until necessary."""
+        apport.crashdb.CrashDatabase.__init__(self, auth_file, options)
+        self.repository_owner = options["repository_owner"]
+        self.repository_name = options["repository_name"]
+        self.app_id = options["github_app_id"]
+        self.labels = set(options["labels"])
+        self.issue_url = None
+        self.github = None
+
+    def _format_report(self, report: apport.Report) -> dict:
+        """Formats report info as markdown and creates Github issue JSON."""
+        body_markdown = ""
+        for key, value in report.items():
+            body_markdown += f"**{key}**\n{value}\n\n"
+
+        return {
+            "title": "Issue submitted via apport",
+            "body": body_markdown,
+            "labels": list(self.labels),
+        }
+
+    def _github_login(self, user_message_callback):
+        with Github(self.app_id, user_message_callback) as github:
+            while not github.authentication_complete():
+                pass
+            return github
+
+    def upload(
+        self,
+        report: apport.Report,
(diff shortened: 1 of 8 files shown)


nteodosio:
Tests will be added once/if the principle is accepted.

bdrung:
The principle concept looks good to me (at least I fail to come up with a complaint :laughing:). So from my side you can continue with test cases and atomic commits.

seb128:
Simon asked for some context. The goal is to allow users to report issues against github projects instead of launchpad using ubuntu-bug which ensure we have the details and log we need. 

While launchpad provide an natural place to report bugs against packages from the archive that's not true for snap. We workarounded that by creating upstream projects on launchpad and redirecting ubuntu-bug there but if the project is actually maintained in github we end up in a situation where the reports go not to the ubuntu package, nor the upstream project but yet another place which is not ideal.

The logic would be for snaps to respect the 'contact' information defined in the snap, if that points to github then direct report the issue there using the github device flow, https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#device-flow. The API key is going to be owned by Canonical IS since the project is under the Canonical org

schopin-pro:
Allow me to run a scenario to see if I understood things correctly.

Let's take the `lxd` snap, which seems to be the only Canonical snap on my system that has a contact URL pointing to GH.

* `ubuntu-bug lxd`, it gathers system info, I get the usual window with the recap of all the data, I click on "send it".
* My browser (let's say Firefox snap) opens to gh.com/lxc/lxd/issues/new with a pre-filled form with all the data that we want
  - If I'm already logged in, it'll be reported under my login?
  - If I'm not logged in it'll be reported as canonical? If so, what can they do with those permissions?

seb128:
> * If I'm not logged in it'll be reported as canonical? If so, what can they do with those permissions?

I'm guessing the question is in reaction to me stating that the API key is owned by Canonical? The key is used to identify the client software, it allows github to identify problematic clients but also to limit the number of requests according to your type of subscription. The bug is submitted by the user using their own account

nteodosio:
I'm having trouble adding tests for `add_snap_info`. This test would require snaps to be installed, but trying [causes an error in the Github CI](https://github.com/nteodosio/apport/actions/runs/3875810090/jobs/6608860480#step:4:1629):

> error: cannot communicate with server: Post "http://localhost/v2/snaps": dial unix /run/snapd.socket: connect: no such file or directory

It seems that [this is a known problem](https://stackoverflow.com/a/55804965).

Any idea on how to work around it?

bdrung:
You can write unit tests and you could write integration/system tests where you mock the snap parts.

nteodosio:
Is github.py supposed to be covered?

bdrung:
Yes,  `github.py` should be covered. `launchpad.py` has test cases which are currently not run in CI. I am working on making them running by using python3-vcr.

nteodosio:
@seb128, was the API key provided yet?

etc/apport/crashdb.conf
```diff
@@ -24,6 +24,21 @@ databases = {
         'recipient': 'submit@bugs.debian.org',
         'sender': ''
     },
+    'snap-github': {
```
bdrung, line 27: Is that GitHub report tied to snaps?

nteodosio: Yes, only for snaps.

apport/crashdb.py
```diff
@@ -777,6 +777,9 @@ def update_traces(self, crash_id, report, comment=""):
             ],
         )
 
+    def external_login(self, ui):
+        """Login to the database via an external resource, e.g. web browser."""
+
```
schopin-pro, line 782: suggestion: remove this API method, fold its content into the  `upload` method instead.

I see the login requirement as an implementation detail of the upload process for this particular backend, not a generic step common to all/most backends.

Furthermore, having it as part of the `upload` method would move it out of the main thread, which makes for a better user experience.

schopin-pro (commented):
Looks fairly good to me, except for the one issue I have with `external_login`.

apport/crashdb.py
nteodosio: Since it requires `ui`, we would have to change the signature of `upload` and all its calls. Is that desirable?

apport/ui.py
```diff
@@ -2011,6 +2017,11 @@ def ui_present_report_details(
             "this function must be overridden by subclasses"
         )
 
+    def ui_github_login(self, text, url, code):
+        return self.ui_info_message(
+            "Login required", text.format(url=url, code=code)
+        )
+
```
schopin-pro, line 2024: issue: this should not be in generic UI code.

The whole point of separating the crash upload into the crashdb interface is to have the rest of the code be blissfully aware of it, having a `ui_github_login` method in the generic UI code sort of flies in the face of it.

apport/crashdb.py
schopin-pro: I'm really not a big fan of having the crashdb implementation driving the UI, since it's supposed to be the other way around.

AFAICT your UI needs boil down to displaying a message to the user, right? If so, how about we "simply" add a new callback to `upload` , e.g. 

```python
    def upload(self, report, progress_callback=None, user_message_callback: Optional[Callable[[], str]] = None):
        pass
```

WDYT?

schopin-pro (changes requested):
I must confess, my previous review was a bit on the light side, because first I'm not familiar with this area of the code, second @bdrung  already had done a review, and finally it was right before lunch.

Now that I have eaten, my previous comment regarding `external_login` and @nteodosio 's answer kind of morphed into a bigger thing.

apport/crashdb.py
nteodosio: That looks like a better solution indeed. @EduardGomezEscandell, is it OK for you?

EduardGomezEscandell: Yeah go ahead. Feel free to make any changes to my code :)

nteodosio:
I addressed the review comments. I added separate commits which I can squash to the corresponding previous ones if this looks good.

schopin-pro (changes requested):
It looks much better!

However, there's still the threading issue to contend with: the `upload` code is in a different thread than the UI, so using `self.ui` in the callback could be a source of issues down the road.

See https://github.com/schopin-pro/apport/tree/wsl where I have stacked up a few commits to change (and add tests). Feel free to pull them in :)

nteodosio:
Thanks! I did so.

bdrung (changes requested):

etc/apport/crashdb.conf
```diff
@@ -24,6 +24,21 @@ databases = {
...
+        'impl': 'github',
+        'repository_owner': None,
+        'repository_name': None,
+        'github_app_id': 'a654870577ad2a2ab5b1', # TODO: Give App ownership to a Canonical github account
```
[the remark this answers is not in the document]
nteodosio: @seb128 might know?

schopin-pro:
@nteodosio FYI, @bdrung asked me yesterday to split off the new upload-time msg API into a separate PR for ease of review, and to basically get the ball rolling. So that's at #123 :)

etc/apport/crashdb.conf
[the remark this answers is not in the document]
seb128: I got a reply from IS, I will follow up on that later today just need to figure out some details

nteodosio:
Alright, once that one is merged I'll rebase.

etc/apport/crashdb.conf
```diff
@@ -24,6 +24,21 @@ databases = {
...
+        'github_app_id': 'bb74ee9268c04aeca4fa',
+        'labels': ['apport'],
+    },
+    'ubuntu-wsl': {
+        'impl': 'github',
+        'repository_owner': 'ubuntu',
+        'repository_name': 'WSL',
+        'github_app_id': 'bb74ee9268c04aeca4fa',
+        'labels': ['apport'],
+        'distro': 'ubuntu wsl'
```
bdrung, line 40: For consistency that should be `Ubuntu WSL` or `Ubuntu on WSL`

nteodosio: Hmm, I fail to see the pattern there. All other `distro:` lines use lowercase. Can you please clarify?

bdrung: I looked at the code again and made a mistake. This `distro` is not the related to `DistroRelease`. This `distro` is the name of the distribution in Launchpad (see doc/crashdb-conf.md). So this `distro` field is useless for the github implementation and can be dropped.

bdrung:
Can you add an entry for the github crash database to `doc/crashdb-conf.md`?

bdrung:
How do you test that? I want to see the code in action before merging.Do you have a test repository to file bugs against?

nteodosio:
> Can you add an entry for the github crash database to doc/crashdb-conf.md?

Sure.

> How do you test that?

Added that to the pull request first post for visibility.

bdrung:
https://github.com/canonical/apport/pull/123 is merged. Can you rebase on main?

nteodosio:
Done, thanks for the heads-up.

schopin-pro:
Please fix the tests? The skip-tests action fails with the following problem:

```
Run python3 -m pytest -ra --cov=$(pwd) --cov-report=xml tests/
============================= test session starts ==============================
platform linux -- Python 3.10.6, pytest-6.2.5, py-1.10.0, pluggy-0.13.0
rootdir: /__w/apport/apport
plugins: cov-3.0.0
collected 488 items / 1 error / 487 selected

==================================== ERRORS ====================================
______________ ERROR collecting tests/integration/test_github.py _______________
ImportError while importing test module '/__w/apport/apport/tests/integration/test_github.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/integration/test_github.py:13: in <module>
    import apport.crashdb_impl.github
apport/crashdb_impl/github.py:16: in <module>
    import requests
E   ModuleNotFoundError: No module named 'requests'

---------- coverage: platform linux, python 3.10.6-final-0 -----------
Coverage XML written to file coverage.xml

=========================== short test summary info ============================
ERROR tests/integration/test_github.py
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 2.59s ===============================
```

apport/crashdb.py
```diff
@@ -769,6 +769,9 @@ def update_traces(self, crash_id, report, comment=""):
             ],
         )
 
+    def external_login(self, user_message_callback):
```
bdrung, line 772: Since we are using a message_callback now, the `external_login` does not need to be part of the public CrashDB.

apport/crashdb_impl/github.py
```diff
@@ -0,0 +1,316 @@
...
+        for key, value in report.items():
+            body_markdown += f"**{key}**\n{value}\n\n"
+
+        return {
+            "title": "Issue submitted via apport",
+            "body": body_markdown,
+            "labels": list(self.labels),
+        }
+
+    def external_login(self, user_message_callback):
```
bdrung, line 177: Since we are using a message_callback now, the external_login does not need to be part of the public CrashDB and it could be named more specific now: `github_login`.

apport/crashdb_impl/github.py
```diff
@@ -0,0 +1,316 @@
...
+        url = "https://github.com/login/device/code"
+        response = self.api_authentication(url, data)
+
+        prompt = "Posting an issue requires a Github account. If you have "
+        prompt += "one, please follow these steps to log in.\n"
+        prompt += "\n"
+        prompt += "Open the following URL. When requested, write this code "
+        prompt += "to enable apport to open an issue.\n"
+        prompt += "URL:  {url}\n"
+        prompt += "Code: {code}"
```
bdrung, line 85:
```suggestion
        prompt = (
            "Posting an issue requires a Github account. If you have "
            "one, please follow these steps to log in.\n"
            "\n"
            "Open the following URL. When requested, write this code "
            "to enable apport to open an issue.\n"
            "URL:  {url}\n"
            "Code: {code}"
        )
```

bdrung: Or even use f-strings:
```suggestion
        prompt = (
            f"Posting an issue requires a Github account. If you have "
            f"one, please follow these steps to log in.\n"
            f"\n"
            f"Open the following URL. When requested, write this code "
            f"to enable apport to open an issue.\n"
            f"URL:  {response['verification_uri']}\n"
            f"Code: {response['user_code']}"
        )
```

bdrung (changes requested):
I'll test this code now against your repository. My plan is to merge the change today or tomorrow and cut a release and upload to Ubuntu to get it in before the feature freeze.

apport/ui.py
```diff
@@ -1819,6 +1822,7 @@ def message_callback(title, text):
             target=self.crashdb.upload,
             args=(self.report, progress_callback, message_callback),
         )
+
```
bdrung, line 1825: One unneeded added newline.

bdrung (approved):

bdrung:
Fix for binary content in some Snap fields: https://github.com/canonical/apport/pull/133

PR #32, opened 2022-12-02
feat: Split bash completions by command

bdrung:
The bash completions directory `/etc/bash_completion.d` is obsolete and the completions should be installed into
`/usr/share/bash-completion/completions`, which has stricter filename requirements.

Therefore move the bash completions into `data/bash-completion` and split each command into a separate file that matches the command name. Keep `apport-bug` and `apport-cli` in one file because they share the same helper functions.

Please review https://github.com/canonical/apport/pull/33 first to fix the missing code coverage for `setup.py`.

### DIFF
--- a/data/bash-completion/apport-bug
+++ b/data/bash-completion/apport-bug
@@ -1,7 +1,4 @@
-# 
-# Apport bash-completion
-#
-###############################################################################
+# Apport bash-completion for apport-bug and apport-cli
 
 # get available symptoms
 _apport_symptoms ()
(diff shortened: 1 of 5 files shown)


schopin-pro (commented):
Dumb question: where in setup.py did it say that the previous completion file was installed in /etc?

bdrung:
`DistUtilsExtra.auto` uses a convention over configuration paradigm and installs the files in `etc/` to `/etc`.

schopin-pro:
Good to know, thanks :)

PR #36, opened 2022-12-05
build: Bump source/target Java version to 7

bdrung:
OpenJDK 17 does not support source/target option 6 any more. The minimum supported version is 7:

```
error: Source option 6 is no longer supported. Use 7 or later.
error: Target option 6 is no longer supported. Use 7 or later.
```

So bump the source/target option to the supported minimum.


bdrung:
I wouldn't backport that change to focal.

PR #37, opened 2022-12-05
test: Test wait_for_proc_cmdline() helper method

bdrung:
The `wait_for_proc_cmdline` helper method can success on the first try. Then the "sleep and retry" part of the function is not used which leads to a reduced code coverage.

Write test cases for the `wait_for_proc_cmdline` helper method to have full code coverage for it.


schopin-pro:
Ugh. Are we really down to testing the test helpers now?!

bdrung:
Test inception. :laughing: Apport already has a bunch of test cases for the test code.

PR #43, opened 2022-12-06
refactor(apport): Avoid duplicate is_same_ns() check

bdrung:
Avoid calling `is_same_ns(host_pid, "mnt")` twice.


schopin-pro:
Good call :)

PR #44, opened 2022-12-06
fix: Fix ignoring SIGXCPU and SIGXFSZ

bdrung:
`str(signal.SIGXCPU)` returns `Signals.SIGXCPU` instead of the string representation of the signal number `24`. Therefore the comparison with the signal number string will never match.


schopin-pro:
Nice catch!

PR #48, opened 2022-12-08
refactor(apport): Split receive_arguments_via_socket out of main

bdrung:
To make the code more readable, split `receive_arguments_via_socket` out of the `main` function.

Add some unit test cases for `receive_arguments_via_socket` to test failure cases (to increase test coverage).


schopin-pro:
LGTM :)

(after scrutinizing it because I though that the diffstat was per commit, not global to the PR, and was wondering where the extra 16 lines came from)

PR #49, opened 2022-12-15
refactor: Add add_tags() and get_tags() to ProblemReport

bdrung:
There are multiple Apport hooks that want to add tags to the report. Add `add_tags` and `get_tags` to `ProblemReport` to avoid code duplication.

### DIFF
--- a/apport/ui.py
+++ b/apport/ui.py
@@ -1558,11 +1558,7 @@ class UserInterface:
 
             # append snap tags, if this report is about the snap
             if "Snap" in self.report and "SnapTags" in self.report:
-                current_tags = self.report.get("Tags", "")
-                if current_tags:
-                    current_tags += " "
-                self.report["Tags"] = current_tags + self.report["SnapTags"]
-                del self.report["SnapTags"]
+                self.report.add_tags(self.report.pop("SnapTags").split(" "))
 
             # show a hint if we cannot auto report a snap bug via 'SnapSource'
             if (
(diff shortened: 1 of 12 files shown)


apport/ui.py
```diff
@@ -1982,10 +1978,7 @@ def add_extra_tags(self):
         """Add extra tags to report specified with --tags on CLI."""
         assert self.report
         if self.args.tag:
-            tags = self.report.get("Tags", "")
-            if tags:
-                tags += " "
-            self.report["Tags"] = tags + " ".join(self.args.tag)
+            self.report.add_tags(self.args.tag)
```
schopin-pro, line 1981: question: the variable is `tag`, singular. Shouldn't it be put in an array instead?

schopin-pro (commented):
Just a small question, otherwise LGTM.

apport/ui.py
bdrung: `args.tag` is a list:
```python
        parser.add_argument(
            "--tag",
            action="append",
            default=[],
```
Should I rename `args.tag` to `args.tags` to reflecting that?

schopin-pro: If that's easily feasible, then yeah I think it'd make sense, but don't got too much out of your way.

bdrung: Let me do it. It should be easy to change.

bdrung: Ready: https://github.com/canonical/apport/pull/73

I will rebase this merge request once PR #73 is reviewed and merged.

PR #50, opened 2022-12-15
feat: Add general hook for collecting cloud build info

bdrung:
The Ubuntu cloud images provide `/etc/cloud/build.info` (generated by livecd-rootfs during image creation time). Example content from https://cloud-images.ubuntu.com/jammy/20221214/jammy-server-cloudimg-amd64-disk-kvm.img:

```
build_name: server
serial: 20221214
```

Parse `/etc/cloud/build.info` if available to tag the report as `cloud-image` and add the content of it to the report.

This merge request depends on https://github.com/canonical/apport/pull/49 which should be reviewed and merged before.

Bug: https://launchpad.net/bugs/1724623


j5awry (approved):
from a cloud perspective, for `build.info`, LGTM

data/general-hooks/image.py
```diff
@@ -0,0 +1,41 @@
...
+            except ValueError:
+                continue
+            if key not in _BUILD_INFO_KEY_MAPPING:
+                continue
+            report[_BUILD_INFO_KEY_MAPPING[key]] = value.strip()
+
+
+def add_info(report, unused_ui):
+    """Add image build information if available."""
+    _add_cloud_build_info(report)
```
schopin-pro, line 41: question: what happens if this raises, e.g. some read error on /etc/cloud/build.info? Does the hook system have a catchall?

schopin-pro (approved):
The concept is sound and the implementation straightforward.

I only have a small question, but it's sort of out of scope, so I'll probably circle back and merge this later today.

data/general-hooks/image.py
bdrung: Yes, `_run_hook` catches `Exception` and adds the crash as `"HookError_" + hookname`.

PR #51, opened 2022-12-16
feat: Move Java crash files into java/testsuite

bdrung:
The Ubuntu package installs `java/crash.class` and `java/crash.jar` into `/usr/share/apport/testsuite/`. Reflect that path by moving `java/crash.java` and the generated artifacts into `java/testsuite`.


schopin-pro:
Small but definite improvement, I like it :)

PR #53, opened 2023-01-06
feat: Unify error message for MalformedProblemReport

bdrung:
Unify error message for `MalformedProblemReport` to always state that the problem report is malformed and ask the user if this a proper .crash text file.


schopin-pro:
That's a good point. We've had multiple crashes related to users running the tool on coredumps instead.

PR #58, opened 2023-01-11
feat: Replace blacklist.d/whitelist.d by report-ignore/report-only directories

bdrung:
The words `blacklist` and `whitelist` may be insensitive. Therefore rename the directory `/etc/apport/blacklist.d` to `/etc/apport/report-ignore` and `/etc/apport/whitelist.d` to `/etc/apport/report-only`. Keep reading the old directories for backward compatibility until Apport 3.0.

This merge request contains the commit from #92: Avoid code duplication in `Report.check_ignored` by factoring out the list files reading into `_read_list_files_in_directory`.

### DIFF
--- a/apport/report.py
+++ b/apport/report.py
@@ -48,9 +48,11 @@ _opt_dir = "/opt"
 # path of the ignore file
 _ignore_file = os.environ.get("APPORT_IGNORE_FILE", "~/.apport-ignore.xml")
 
-# system-wide blacklist
-_blacklist_dir = "/etc/apport/blacklist.d"
-_whitelist_dir = "/etc/apport/whitelist.d"
+# system-wide denylist/allowlist directories
+_DENYLIST_DIR = "/etc/apport/report-ignore"
+_ALLOWLIST_DIR = "/etc/apport/report-only"
+_LEGACY_DENYLIST_DIR = "/etc/apport/blacklist.d"  # wokeignore:rule=blacklist
+_LEGACY_ALLOWLIST_DIR = "/etc/apport/whitelist.d"  # wokeignore:rule=whitelist
 
 # programs that we consider interpreters
 interpreters = [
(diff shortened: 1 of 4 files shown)


schopin-pro:
Code-wise LGTM, but thinking a bit more about deny vs ignore, I'm actually reverting my opinion back to ignore. The issue with `deny` is: what are you denying? "Ignore" seems fairly straightforward, you simply don't do anything related to this binary.

I can understand it as "I deny this binary the right to crash, so you MUST report it", which is basically the semantics of the current "whitelist", or "I deny this binary the right to report the crashes".

OTOH, we have the same issue with allow, and I don't have a good word for it. So maybe we could just have "report-deny" and "report-allow" to make it clearer what we're denying/allowing?

bdrung:
Thanks for the feedback. Ideas: ignore/report or ignore/process?

ogayot:
Sounds a bit silly but I would suggest `ignore` and `no-ignore`. Do we know what's the expected behavior when a pattern is present in both the blacklist and whitelist?

bdrung:
"If something is defined in /etc/apport/whitelist.d/, everything which does not match gets ignored then."

waveform80:
Perhaps "include/exclude"? The context is essentially filtration after all.

> OTOH, we have the same issue with allow, and I don't have a good word for it. So maybe we could just have "report-deny" and "report-allow" to make it clearer what we're denying/allowing?

This sounds like an excellent suggestion to me. It's a bit more wordy, but explicit is better than implicit; it also makes the filtration context more obvious. Thumbs up for "fix it with more words!"

daniloegea:
include/exclude or include/ignore sounds good enough.

clang-format uses `.clang-format-include` and `.clang-format-ignore`.

woke itself uses `ignore_files`

bdrung:
So we have some options:
1. report-deny / report-deny
2. report-ignore / report-only
3. programs-ignore / programs-only

bdrung:
If no-one objects, I will change it to option 2: report-ignore / report-only.

schopin-pro:
Ooooh I really like "report-only" :)

PR #64, opened 2023-01-12
test: Skip tests when required binaries are missing

bdrung:
Some test cases need some binaries for their test which might not be installed. Check for the presence and skip if the binaries are missing.


bdrung:
The newly added code for skipping test when required binaries are missing is not covered. So this commit needs to be bundled with the commit that adds CI tests for it: https://github.com/canonical/apport/pull/87

PR #65, opened 2023-01-12
test: Skip tests if Python dependencies are missing

bdrung:
The Python crashes system tests need `dbus`, the GTK UI system tests need GI and the Glib/GTK libraries, and
`test_receive_arguments_via_socket_invalid_socket` needs the systemd library. Check for the presence and skip if the libraries are missing.

### DIFF
--- a/tests/system/test_python_crashes.py
+++ b/tests/system/test_python_crashes.py
@@ -18,13 +18,19 @@ import tempfile
 import textwrap
 import unittest
 
-import dbus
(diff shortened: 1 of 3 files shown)


bdrung:
The newly added code for skipping test when Python dependencies are missing is not covered. So this commit needs to be bundled with the commit that adds CI tests for it: https://github.com/canonical/apport/pull/87

PR #66, opened 2023-01-12
fix: Replace deprecated distutils for report test

bdrung:
The test case `test_check_interpreted` triggers a deprecation warning: "The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives"

`Report._python_module_path` uses `importlib.util.find_spec` which tries to find the spec for the module (but does not execute the module).

Replace `distutils.cmd` by `http.server` in the test, which is a real world example:

```
python3 -m http.server 8080
```

### DIFF
--- a/tests/integration/test_report.py
+++ b/tests/integration/test_report.py
@@ -502,10 +502,10 @@ class T(unittest.TestCase):
             pr = apport.report.Report()
             pr["ExecutablePath"] = "/usr/bin/python3"
             pr["ProcStatus"] = "Name:\tpython3"
-            pr["ProcCmdline"] = "python\0-m\0distutils.cmd\0foo"
+            pr["ProcCmdline"] = "python3\0-m\0http.server\08080"
             pr._check_interpreted()
             self.assertEqual(pr["InterpreterPath"], "/usr/bin/python3")
-            self.assertIn("distutils/cmd.py", pr["ExecutablePath"])
+            self.assertIn("http/server.py", pr["ExecutablePath"])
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_report.py
```diff
@@ -502,10 +502,10 @@ def test_check_interpreted(self):
             pr = apport.report.Report()
             pr["ExecutablePath"] = "/usr/bin/python3"
             pr["ProcStatus"] = "Name:\tpython3"
-            pr["ProcCmdline"] = "python\0-m\0distutils.cmd\0foo"
+            pr["ProcCmdline"] = "python3\0-m\0http.server\08080"
```
schopin-pro, line 505: This isn't really eye-for-eye replacement, the distutils command is one-and-done whereas http.server is long running.

bdrung: This integration test is not running this command, it is only checking if the command line is an interpreter.

schopin-pro: If it's not running the command, how come we have a warning? :thinking:

bdrung: `_python_module_path` uses `importlib.util.find_spec`.

bdrung: I added following sentence to the commit message: "`Report._python_module_path` uses `importlib.util.find_spec` which tries to find the spec for the module (but does not execute the module)."

schopin-pro:
Much better with the new details in the commit log, thanks :)

PR #77, opened 2023-01-18
refactor: Introduce apport.ui.Action dataclass

bdrung:
Use a dataclass as response for `UserInterface.ui_present_report_details` instead of a dictionary. This ensures that the attributes are named correctly and allows specifying default values for the attributes.


schopin-pro (approved):
Great stuff, I'm loving those diffstats!

PR #78, opened 2023-01-18
fix: Import needed packaging library directly

bdrung:
`apport.crashdb` imports `apport.report`, but only uses `apport.packaging` (which is declared in `apport/__init__.py`).

Import `packaging` directly from `apport.packaging_impl` instead. This change is needed to avoid circular imports when fixing https://launchpad.net/bugs/2003098.

### DIFF
--- a/apport/crashdb.py
+++ b/apport/crashdb.py
@@ -17,7 +17,7 @@ import urllib.error
 import urllib.parse
 import urllib.request
 
-import apport.report
+from apport.packaging_impl import impl as packaging
 
 
 def _u(string):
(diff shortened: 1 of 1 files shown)


bdrung:
Test coverge improvement: https://github.com/canonical/apport/pull/81

schopin-pro:
I'm actually surprised it worked before this patch. I assumed that `import apport.report ` wouldn't make the toplevel `apport` namespace available fully, but apparently I was wrong.

Live and learn :)

(bypassing the codecov thing)

PR #83, opened 2023-01-20
refactor(apport): Use logging library for logging

bdrung:
To make the code better readable, use the standard `logging` library for logging in `data/apport`.

### DIFF
--- a/data/apport
+++ b/data/apport
@@ -26,6 +26,7 @@ import fcntl
 import grp
 import inspect
 import io
+import logging
 import os
 import pwd
 import re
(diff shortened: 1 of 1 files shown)


data/apport
```diff
@@ -179,22 +181,19 @@ def init_error_log():
         sys.stderr = io.TextIOWrapper(sys.stderr)
         sys.stdout = sys.stderr
 
-
-def error_log(msg):
-    """Output something to the error log."""
-
-    apport.error("apport (pid %s) %s: %s", os.getpid(), time.asctime(), msg)
+    logging.basicConfig(format=LOG_FORMAT)
```
schopin-pro, line 184: question: the function could do an early return, should we take it into account?

schopin-pro (approved):
Overall LGTM but I have a slight concern around logger initialization.

data/apport
bdrung: Good catch. I moved the logging configuration into the main function.

schopin-pro (approved):

bdrung:
Ignoring failed `codecov/patch` in this case.

PR #87, opened 2023-01-20
test: Skip tests if Python dependencies or required binaries are missing

bdrung:
The Python crashes system tests need `dbus`, the GTK UI system tests need GI and the Glib/GTK libraries, and `test_receive_arguments_via_socket_invalid_socket` needs the systemd library. Check for the presence and skip if the libraries are missing.

Some test cases need some binaries for their test which might not be installed. Check for the presence and skip if the binaries are missing.

Run all tests in a minimal container to check if they are skipped or succeed.

### DIFF
--- a/tests/integration/test_packaging_apt_dpkg.py
+++ b/tests/integration/test_packaging_apt_dpkg.py
@@ -546,6 +546,7 @@ deb http://secondary.mirror tuxy extra
         self.assertNotEqual(arch, "")
         self.assertNotIn("\n", arch)
 
+    @skip_if_command_is_missing("dpkg-architecture")
     def test_get_library_paths(self):
         """get_library_paths()."""
         paths = impl.get_library_paths()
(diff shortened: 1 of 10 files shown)


tests/integration/test_packaging_apt_dpkg.py
```diff
@@ -546,6 +546,7 @@ def test_get_system_architecture(self):
         self.assertNotEqual(arch, "")
         self.assertNotIn("\n", arch)
 
+    @skip_if_command_is_missing("dpkg-architecture")
```
schopin-pro, line 549: nitpick (not-blocking): there's a consistency issue, some are full paths, other are just basenames

schopin-pro (approved):
LGTM

(single non-blocking very nitpicky nitpick)

tests/integration/test_packaging_apt_dpkg.py
bdrung: Only one uses a full path (`/usr/bin/sensible-pager`) and that's because apport-cli calls the binary with the full path.

PR #89, opened 2023-01-20
refactor(apport): Use f-strings

bdrung:
The use of f-strings is preferred instead of using the `%` formatting operator.


schopin-pro (approved):

bdrung:
Ignoring codecov/patch and merging anyway.

PR #90, opened 2023-01-20
test: Add unit tests for is_closing_session()

bdrung:
Checking for root processes does not belong into `is_closing_session`.

Add unit tests for `is_closing_session()` to increase code coverage to ease future refactoring.

### DIFF
--- a/data/apport
+++ b/data/apport
@@ -363,7 +363,7 @@ def _run_with_output_limit_and_timeout(
     return stdout, stderr
 
 
-def is_closing_session():
+def is_closing_session() -> bool:
     """Check if pid is in a closing user session.
 
     During that, crashes are common as the session D-BUS and X.org are going
--- a/data/apport
+++ b/data/apport
@@ -369,10 +369,6 @@ def is_closing_session():
     During that, crashes are common as the session D-BUS and X.org are going
     away, etc. These crash reports are mostly noise, so should be ignored.
     """
-    # consistency check, don't do anything for root processes
-    if crash_uid == 0 or crash_gid == 0:
-        return False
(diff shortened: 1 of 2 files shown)


schopin-pro:
I'm guessing it conflicts with the `woke` renaming work, given the negative diff contains "sanity"?

data/apport
```diff
@@ -1031,7 +1027,8 @@ def process_crash(options: argparse.Namespace) -> int:
     # write out the user coredumps
     recover_privileges()
 
-    if is_closing_session():
+    # Do not check closing session for root processes
+    if crash_uid != 0 and crash_gid != 0 and is_closing_session():
```
schopin-pro, line 1031: question: why do we want to special-case root here?

I mean I have a vague idea why, but that's somewhat fuzzy.

schopin-pro (approved):
LGTM, I only have a question that's only tangentially related to the PR's purpose.

data/apport
bdrung: Isn't "Checking for root processes does not belong into `is_closing_session`." explanition enough?

schopin-pro: I mean I understand why you moved it out, it makes sense, I'm more wondering why we had the check in the first place. As I said, only tangentially related ;).

I'll merge anyway.

bdrung: Let me search in the git history and improve the comment.

bdrung: It was introduced in commit e1c90227666d9b0f95c3cd6229df1f9097974f4f to fix CVE-2022-28655. It still does not explain the full reason.

PR #93, opened 2023-01-25
Move snap contact info parser into a function and test it

nteodosio:


apport/report.py
```diff
@@ -447,19 +447,28 @@ def add_snap_info(self, snap):
...
-                r"((?:[^\/]+\/\+source\/)?[^\/]+)(?:.*field\.tags?=([^&]+))?"
-            )
-            m = re.search(p, urllib.parse.unquote(snap.get("contact", "")))
-            if m and m.group(1):
-                self["SnapSource"] = m.group(1)
-                if m.group(2):
-                    self["SnapTags"] = m.group(2)
+            self._add_snap_contact_info(snap)
+
+    def _add_snap_contact_info(self, snap):
```
bdrung, line 452: Can you add type annotation to it?

apport/report.py
```diff
@@ -447,19 +447,28 @@ def add_snap_info(self, snap):
...
+        # Parse project (e.g. 'subiquity') or source package
+        # string (e.g. 'ubuntu/+source/gnome-calculator') from snap
+        # 'contact'. Additionaly, extract any tag/tags defined in the
+        # contact URL.
```
bdrung, line 456: This comment could be docstring.

apport/report.py
```diff
@@ -447,19 +447,28 @@ def add_snap_info(self, snap):
...
+        p = (
+            r"(^https?:\/\/.*launchpad\.net\/"
+            r"((?:[^\/]+\/\+source\/)?[^\/]+)(?:.*field\.tags?=([^&]+))?)|"
+            r"(^https?://.*github\.com/([^/]+)/([^/]+))"
```
bdrung, line 460: This belongs into a separate commit, since that's more than refactoring.

tests/unit/test_report.py
```diff
@@ -1083,3 +1083,23 @@ def test_suspend_resume(self):
...
+
+    def test__add_snap_contact_info(self):
+        """_add_snap_contact_info()."""
+        # test that report is against Launchpad
+        pr = apport.report.Report()
+        snap = {
+            "contact": "https://bugs.launchpad.net/ubuntu/+source/"
+            "chromium-browser/+bugs?field.tag=snap"
+        }
+        pr._add_snap_contact_info(snap)
```
bdrung, line 1095: Private functions should not be tested directly. Either test it through its caller or make it public. I tend to prefer the former in this case.

apport/report.py
bdrung, line 452: The function could take `snap_contact: str` instead of the complete `snap`.

bdrung (changes requested):

nteodosio:
Hmm I suppose that since now the testing is via the caller it can't be a unit test anymore?

bdrung:
IMO it's still a unit test. It would need to mock the snap calls.

nteodosio:
Yes just realized that in the last push (:

nteodosio:
There were too many things to mock along the way (including the yaml import).

I think it's better to make it not private instead.

tests/unit/test_report.py
```diff
@@ -1083,3 +1083,23 @@ def test_suspend_resume(self):
...
+        pr = apport.report.Report()
+        snap = (
+            "https://bugs.launchpad.net/ubuntu/+source/"
+            "chromium-browser/+bugs?field.tag=snap"
+        )
+        pr.add_snap_contact_info(snap)
+        self.assertEqual("ubuntu/+source/chromium-browser", pr["SnapSource"])
+        self.assertEqual("snap", pr["SnapTags"])
+
+        # Test that report is against Github
```
bdrung, line 1099: suggestion: This could be a separate test method, e.g. `test_add_snap_contact_info_github`. The other part could be named `test_add_snap_contact_info_launchpad`.

tests/unit/test_report.py
```diff
@@ -1083,3 +1083,15 @@ def test_suspend_resume(self):
         del pr["dmi.bios.version"]
         expected = "suspend/resume:Cray XT5"
         self.assertEqual(expected, pr.crash_signature())
+
+    def test_add_snap_contact_info_launchpad(self):
+        """add_snap_contact_info()"""
+        # Test that report is against Launchpad
```
bdrung, line 1089: nitpick: That comment can be merged into the docstring now.

PR #94, opened 2023-01-26
Parse Github URL from snap contact fields and test it

nteodosio:
Split out from #93.

### DIFF
--- a/apport/report.py
+++ b/apport/report.py
@@ -456,6 +456,7 @@ class Report(problem_report.ProblemReport):
         (e.g. 'ubuntu/+source/gnome-calculator') from snap 'contact'.
         Additionaly, extract any tag/tags defined in the contact URL.
         """
+        # Launchpad
         p = (
             r"^https?:\/\/.*launchpad\.net\/"
             r"((?:[^\/]+\/\+source\/)?[^\/]+)(?:.*field\.tags?=([^&]+))?"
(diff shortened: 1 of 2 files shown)


apport/report.py
```diff
@@ -457,14 +457,20 @@ def add_snap_contact_info(self, snap_contact: str) -> None:
         Additionaly, extract any tag/tags defined in the contact URL.
         """
         p = (
-            r"^https?:\/\/.*launchpad\.net\/"
-            r"((?:[^\/]+\/\+source\/)?[^\/]+)(?:.*field\.tags?=([^&]+))?"
+            r"(^https?:\/\/.*launchpad\.net\/"
+            r"((?:[^\/]+\/\+source\/)?[^\/]+)(?:.*field\.tags?=([^&]+))?)|"
+            r"(^https?://.*github\.com/([^/]+)/([^/]+))"
```
bdrung, line 462: Since these two matches share nothing, the code will probably more readable to search for them separately.

apport/report.py
```diff
@@ -457,14 +457,20 @@ def add_snap_contact_info(self, snap_contact: str) -> None:
...
-                self["SnapTags"] = m.group(2)
+        if m:
+            if m.group(1):  # Launchpad
+                self["SnapSource"] = m.group(2)
+                if m.group(3):
+                    self["SnapTags"] = m.group(3)
+            elif m.group(4):  # Github
+                self["SnapGitOwner"] = m.group(5)
+                self["SnapGitName"] = m.group(6)
+                self["CrashDB"] = "snap-github"
```
bdrung, line 473: I just noticed: This change requires the CrashDB `snap-github`. So it cannot go alone.

nteodosio: Indeed. Is there something to do about this on my side or should we wait for #26?

bdrung: The commits can go back into #26.

PR #96, opened 2023-01-27
refactor: Simplify cmp() in CrashDatabase._duplicate_search_signature

bdrung:
`cmp()` in `CrashDatabase._duplicate_search_signature()` has more if-conditions than return values. Group them by return value:

| x    | y    | return |
|------|------|--------|
| None | None | 0      |
| ""   | ""   | 0      |
| None | ""   | 1      |
| None | .*   | 1      |
| ""   | .*   | 1      |
| ""   | None | -1     |
| .*   | None | -1     |
| .*   | ""   | -1     |

### DIFF
--- a/apport/crashdb.py
+++ b/apport/crashdb.py
@@ -508,17 +508,9 @@ class CrashDatabase:
             y = y[1]
             if x == y:
                 return 0
-            if x == "":
-                if y is None:
-                    return -1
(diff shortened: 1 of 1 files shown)


schopin-pro (approved):
LGTM. While still a bit of a head-scratcher, the logic is a bit easier to understand, and it's shorter, too.

PR #99, opened 2023-02-02
refactor(apport): Return early in init_error_log

bdrung:
Return early in `init_error_log` to reduce the indentation level and make the function more readable.


schopin-pro (approved):
Should we rebase this to pick up the new test?

bdrung:
That was the plan. Rebased.

PR #100, opened 2023-02-02
refactor: Introduce get_process_environ

bdrung:
Reduce duplicate code by introducing `apport.fileutils.get_process_environ` which reads and parses the process environ. Also make `Report.add_proc_environ` more readable by move some logic into a new `_Environment` object.


schopin-pro (approved):
I like this, makes the `add_proc_environ` function much more readable!

PR #101, opened 2023-02-02
fix(whoopsie-upload-all): Use NoninteractiveHookUI for add_hooks_info

bdrung:
Commit f8edc7280d492e710218c193e53542f1f86cfa91 introduces a regression: The package specific hooks are not run in `whoopsie-upload-all` because `add_hooks_info` is called with `ui=None`. `add_hooks_info` will raise an `AssertionError` for `ui=None`.

Use `NoninteractiveHookUI()` when `ui` is set to `None`.

Bug: https://launchpad.net/bugs/2003098

### DIFF
--- a/apport/report.py
+++ b/apport/report.py
@@ -39,6 +39,7 @@ import apport.logging
 import problem_report
 from apport.hookutils import kill_pkttyagent
 from apport.packaging_impl import impl as packaging
+from apport.ui import HookUI, NoninteractiveHookUI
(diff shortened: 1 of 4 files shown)


schopin-pro (approved):
I had concerns about potentially changing the API contract for external hooks, but an out-of-band discussion with @bdrung alleviated the concerns, as returning None already has an ascribed semantic.

PR #105, opened 2023-02-03
refactor(apport): Avoid consecutive sys.stderr assignment

bdrung:
Mypy complains about incompatible types:

```
$ mypy data/apport
data/apport:180: error: Incompatible types in assignment (expression has type "BufferedWriter", variable has type "TextIO")
data/apport:181: error: Argument 1 to "TextIOWrapper" has incompatible type "TextIO"; expected "IO[bytes]"
[...]
```


bdrung:
I think that `codecov/project` complains wrongly.

schopin-pro (approved):

PR #107, opened 2023-02-03
fix: Close socket in forward_crash_to_container

bdrung:
The socket in `forward_crash_to_container` is not closed.


schopin-pro:
Bypassing branch protection because it's only codecov being a pain.

PR #110, opened 2023-02-08
fix: Only sent UNIX credentials when having CAP_SYS_ADMIN

bdrung:
Sending `SCM_CREDENTIALS` needs the capability `CAP_SYS_ADMIN` to send another process ID than its own. Since the sent `host_pid` differs from the apport process ID, `CAP_SYS_ADMIN` is needed.

Checking for `CAP_SYS_ADMIN` need the prctl() syscall which is currently only available in `python3-prctl`. Checking `os.geteuid() == 0` is not enough. So just allow specifying `has_cap_sys_admin` to `forward_crash_to_container`.

Convert `array.array` to `bytes` to make the `ancillary` consistently of type `list[typing.Tuple[int, int, bytes]]` to make mypy happy.

### DIFF
--- a/data/apport
+++ b/data/apport
@@ -510,7 +510,7 @@ def is_same_ns(pid, ns):
 
 
 def forward_crash_to_container(
-    host_pid: int, options: argparse.Namespace
+    host_pid: int, options: argparse.Namespace, has_cap_sys_admin: bool = True
 ) -> None:
     """Try to forward the crash to the container.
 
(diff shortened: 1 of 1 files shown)


schopin-pro:
I'm confused, you never update the call site? Also, this needs rebasing :)

bdrung:
Since codecov is not happy, lets bundle this commit with the commit that adds the test case: https://github.com/canonical/apport/pull/113

PR #112, opened 2023-02-08
refactor(apport): Drop unused cwd

bdrung:
The core files are no longer written to the current working directory of the crashed process, but into `/var/lib/apport/coredump`. `dir_fd` is only taken into account if the specified directory is relative. Since the coredump directory is absolute, `dir_fd` and therefore `cwd` are not used.

### DIFF
--- a/data/apport
+++ b/data/apport
@@ -79,13 +79,7 @@ def check_lock():
         signal.signal(signal.SIGALRM, original_handler)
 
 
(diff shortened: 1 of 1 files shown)


schopin-pro (commented):
Are we sure about that one? Because core paths *can* be relative, according to https://github.com/canonical/apport/blob/main/apport/fileutils.py#L30 since the environment variable can be , say, `"./cores"`

PR #114, opened 2023-02-09
refactor(apport): Replace global proc_pid_fd by ProcPid

bdrung:
Replace global variable `proc_pid_fd` by an `ProcPid` instance that is a context manager to ensure the opened file will be closed.

### DIFF
--- a/data/apport
+++ b/data/apport
@@ -49,6 +49,31 @@ import apport.report
 LOG_FORMAT = "%(levelname)s: apport (pid %(process)s) %(asctime)s: %(message)s"
 
 
+class ProcPid(contextlib.ContextDecorator):
+    """Context manager to access /proc/<pid>."""
+
+    def __init__(self, pid: int, path: typing.Optional[str] = None) -> None:
+        self.path = path or f"/proc/{pid}"
+        self.fd: typing.Optional[int] = None
+
+    def __enter__(self):
+        self.fd = os.open(self.path, os.O_RDONLY | os.O_PATH | os.O_DIRECTORY)
+        return self
+
+    def __exit__(self, *exc):
+        if self.fd is not None:
+            os.close(self.fd)
+        return False
+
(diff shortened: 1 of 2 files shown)


bdrung:
IMO we can ignore codecov/project since the merge reduces the amount of code.

data/apport
```diff
@@ -49,6 +49,22 @@ import apport.report
 LOG_FORMAT = "%(levelname)s: apport (pid %(process)s) %(asctime)s: %(message)s"
 
 
+class ProcPid:
+    """Access /proc/<pid>."""
+
+    def __init__(self, pid: int, path: typing.Optional[str] = None) -> None:
+        self.fd = os.open(
+            path or f"/proc/{pid}", os.O_RDONLY | os.O_PATH | os.O_DIRECTORY
+        )
```
schopin-pro, line 58: issue (design): should we really embed a FD in a plain class?

This pattern means we're relying on the finalizer closing out the FD, which sounds like a bad idea to me.

schopin-pro (changes requested):
I have a design issue that I'd like addressed.

data/apport
bdrung: Very good point. It will be probably much better to convert that class into a context manager.

schopin-pro (commented):
question: doesn't that open us to race condition?

Disclaimer: I'm really not well-versed in the particulars of procfs, crash-handling, and all the low-level notions here, so I might be raising moot points.

In the previous design, it's fairly obvious that we only open the PID directory once, which I'd assume protects us in case of PID recycling by the kernel (can it recycle the PID of a process that just crashed?). Do we still have that "atomic" property with the new one? I *think* we do, but I can't figure out the call tree well enough to be sure.

bdrung:
It will not open us to race conditions. We create only one `ProcPid` object for the relevant PID.

PR #115, opened 2023-02-10
refactor(apport): Introduce refine_core_ulimit

bdrung:
Carve `refine_core_ulimit` out of `process_crash` to make that function smaller.


schopin-pro (commented):
Slight naming issue :)

data/apport
```diff
@@ -927,24 +948,11 @@ def process_crash(options: argparse.Namespace) -> int:
...
-        )
-        core_ulimit = -1
-
-    if options.dump_mode == 2:
-        logger.error(
-            "not creating core for pid with dump mode of %s", options.dump_mode
-        )
-        # a report should be created but not a core file
-        core_ulimit = 0
+    core_ulimit = determine_core_ulimit(options)
```
schopin-pro, line 955: issue: reading this function doesn't make sense, we already have a "options.core_ulimit", and then we try to determine it?

Maybe "check", "refine" instead?

(since this is a refactor PR, I consider legibility to be primary concern)

schopin-pro (commented):
Slight naming issue

(take two)

data/apport
bdrung: Agreed. "refine" sounds like a better name. I'll change it.

schopin-pro (approved):

PR #122, opened 2023-02-15
test: Specify Python files for linters directly

bdrung:
Black and pylint might complain about Python files in `.pc` that were created by quilt containing pre-patch versions of the Python files. Pylint might also check files in `.pybuild`. All these files should not be checked by the linters.

### DIFF
--- a/tests/run-linters
+++ b/tests/run-linters
@@ -14,18 +14,19 @@
 
 set -eu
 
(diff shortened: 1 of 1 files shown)


schopin-pro (approved):
*Someone* has been working in a packaging tree... :grin:

PR #123, opened 2023-02-16
ui: upload-time messaging

schopin-pro:
Some crashdb backends might need to display messages to the user, here's a safe API to do so.

See individual commit logs for more detailed descriptions of the changes.

This is an offshoot of #26

### DIFF
--- a/apport/ui.py
+++ b/apport/ui.py
@@ -24,6 +24,7 @@ import gzip
 import io
 import locale
 import os.path
+import queue
 import re
 import shutil
 import signal
@@ -1801,6 +1802,13 @@ class UserInterface:
         def progress_callback(sent, total):
             self.upload_progress = float(sent) / total
 
+        message_queue = queue.SimpleQueue()
+
+        def message_callback(title, text):
+            message_displayed = threading.Event()
+            message_queue.put((title, text, message_displayed))
+            message_displayed.wait()
+
         # drop internal/uninteresting keys, that start with "_"
         for k in list(self.report):
             if k.startswith("_"):
@@ -1808,14 +1816,21 @@ class UserInterface:
 
         self.ui_start_upload_progress()
         upthread = apport.REThread.REThread(
-            target=self.crashdb.upload, args=(self.report, progress_callback)
+            target=self.crashdb.upload,
+            args=(self.report, progress_callback, message_callback),
         )
         upthread.start()
         while upthread.is_alive():
             self.ui_set_upload_progress(self.upload_progress)
             try:
-                upthread.join(0.1)
+                title, text, msg_displayed = message_queue.get(
+                    block=True, timeout=0.1
+                )
+                self.ui_info_message(title, text)
+                msg_displayed.set()
                 upthread.exc_raise()
+            except queue.Empty:
+                pass
             except KeyboardInterrupt:
                 sys.exit(1)
             except (smtplib.SMTPConnectError, urllib.error.URLError) as error:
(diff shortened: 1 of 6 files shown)


schopin-pro:
V2 patchset changes:
* re-ordering of commits
* formatting
* fold the get_nowait() + thread.join() into a blocking get()
* add missing changes on crashdb implementations. Whoops!

schopin-pro:
... and a V3 to fix bisectability. Sorry about that :)

tests/integration/test_ui.py
```diff
@@ -636,10 +660,14 @@ def test_run_report_bug_package(self):
         """run_report_bug() for a package"""
         self.ui = UserInterfaceMock(["ui-test", "-f", "-p", "bash"])
         self.ui.present_details_response = apport.ui.Action(report=True)
+        self.ui.crashdb.upload_delay = 0.2
```
bdrung, line 663: Shouldn't this current version work without an upload delay?

tests/integration/test_ui.py
```diff
@@ -32,6 +32,30 @@
...
+        self.upload_delay = 0.0
+        self.upload_msg_title = ""
+        self.upload_msg_text = ""
+
+        super().__init__(*args, **kwargs)
+
+    def upload(
+        self, report, progress_callback=None, user_message_callback=None
+    ):
+        if user_message_callback and self.upload_msg_text:
```
bdrung, line 46: Can you move this logic into  `apport.crashdb_impl.memory.CrashDatabase`? This implementation is designed for testing.

bdrung (changes requested):

tests/integration/test_ui.py
schopin-pro: Yes it should, the delay was introduced to first make sure the new feature did not break the current common case (i.e. empty polling without messaging).

I can remove it if you prefer, it shouldn't change much.

tests/integration/test_ui.py
schopin-pro: It's not just designed for testing though, as it's used in `bin/dupdb-admin`, hence my reluctance to modify it directly.

tests/integration/test_ui.py
bdrung: Yes, please. Having random delays is normally an indication for possible race conditions.

tests/integration/test_ui.py
bdrung: I am undecided where the best place is. The memory `CrashDatabase` has already an `sample_data` option which is clearly only for testing.

schopin-pro: Moved the feature in the memory database directly.

tests/integration/test_ui.py
schopin-pro: I reimplemented the tests entirely in 3 separate tests rather than hogging an already existing test. One has no delay, no msg, one has delays, one has msgs.

apport/ui.py
```diff
@@ -1801,21 +1802,35 @@ def file_report(self):
         def progress_callback(sent, total):
             self.upload_progress = float(sent) / total
 
+        message_queue = queue.SimpleQueue()
+
+        def message_callback(title, text):
+            message_displayed = threading.Event()
+            message_queue.put((title, text, message_displayed))
```
bdrung, line 1809: suggestion: Instead of using a tuple, a dataclass could be used.

bdrung: Merging anyway.

PR #124, opened 2023-02-17
test: Increase waiting timeout to 15 seconds

bdrung:
The system tests `test_write_reports` failed on GitHub CI on the Ubuntu 22.04 host with AssertionError: Process /usr/bin/sleep not started within 5 seconds.

Increase the waiting timeout to 15 seconds.


bdrung:
Increasing the timeout does not help. There must be some other underlying problem.

PR #127, opened 2023-02-23
test: Assert _run_test_executable runner is alive

bdrung:
Sometimes `test_write_reports` fails in GitHub CI with AssertionError: Process /usr/bin/sleep not started within 5 seconds. In that case the runner was not started. Add assertions for future cases.


schopin-pro (approved):

bdrung:
Gladly this merge request failed (see https://github.com/canonical/apport/actions/runs/4252159249/jobs/7395822886) which indicates that the problem lies somewhere else and these assertions are not needed.

PR #129, opened 2023-02-23
test: Move unkillable_shutdown system tests to integration tests

bdrung:
Commit 1034142e3379d00443ad7b5cd6b0e0aa48de1dd1 ("refactor: Make unkillable_shutdown system tests run in containers") removed the requirement on systemd for running the unkillable_shutdown tests. Therefore these tests can be moved back to integration tests.


bdrung:
Please ignore the reduced code coverage because `FileExistsError` is not covered any more.

PR #130, opened 2023-02-23
test: Communicate test process ID via a queue

bdrung:
Instead waiting for the test process to appear, communicate the test process ID via a queue. This requires less code.


schopin-pro (approved):
Nice one, should be less brittle!

PR #131, opened 2023-02-23
feat: Already add package info during crash

bdrung:
The package info is added to the crash report when the report is reported by the user or when it is uploaded. There can be quite some time delay and the list of installed packages could have change since then. There are crashes on errors.ubuntu.com where the reported version is newer than the code in the Python stack trace. So this kind of discrepancy happens in the real world.

So try to add the package info during crash. Add a `HooksRun` entry to the report to document that the hooks were not run (for `whoopsie-upload-all`). Change `add_package_info` to not overwrite existing entries.

### DIFF
--- a/apport/report.py
+++ b/apport/report.py
@@ -385,6 +385,29 @@ class Report(problem_report.ProblemReport):
 
         return version
 
+    def _get_transitive_dependencies(self, package: str) -> str:
+        # get set of all transitive dependencies
+        dependencies_set: set[str] = set()
+        _transitive_dependencies(package, dependencies_set)
+
+        # get dependency versions
+        dependencies = ""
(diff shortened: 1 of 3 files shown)


schopin-pro (approved):
I don't see anything really wrong with this, so it's mergeable.

However, I'm confused about the HooksRun thing. Why do we need it? Its semantics are weird, because its only value is "no", and we only check the field's existence, not its value, so if somewhere along the line some other code changes it to "yes" because it wants to say that the hooks *have* been run, the behavior stays the same.

schopin-pro:
For the record, this seems to have caused a regression, as some code in UI relied on the presence of the Dependencies as meaning that the crash file is "complete".

PR #139, opened 2023-02-24
tests: skip the GH upload test if internet isn't available

schopin-pro:
Typically, this would be the case when building a Debian or Ubuntu package.

Another approach would have been to mock out completely the function that reaches out to GH, but I figure it's a good idea to have it actually run regularly to ensure we quickly detect issues such as expiring tokens.

### DIFF
--- a/tests/integration/test_github.py
+++ b/tests/integration/test_github.py
@@ -9,6 +9,7 @@
 
 import unittest
 from unittest.mock import ANY, Mock, patch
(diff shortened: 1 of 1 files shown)


bdrung:
I tend to prefer to mock this integration test. We could add a system test that queries GitHub.

bdrung:
Superseded by https://github.com/canonical/apport/pull/140 which got merged.

PR #140, opened 2023-02-24
fix: Mock query to Github in integration test.

nteodosio:
Autopkgtests fail otherwise because they cannot reach Github.


bdrung:
Thanks. The change looks good and the apport package builds fine now. I will upload apport with your patch.

Two smaller nitpicks and one request:
* Can you remove the dot from the commit title?
* It's not autopkgtests that fails, but running the tests during package build. Only during package build the Internet access is blocked.
* Your change reduces the code coverage. Can you add some more tests for the now untested parts? Note: There are mocking libraries for mocking network traffic.

nteodosio:
I added a system test but Codecov is considering the test itself as uncovered.

PR #148, opened 2023-03-27
fix: parsing options in sources.list (LP: #1822712)

xypron:
In sources.list the enclosing brackets of the options are separated by spaces from option values, e.g.

    deb [ arch=riscv64 ] http://ports.ubuntu.com/ubuntu-ports/ lunar main

Fixes: f7ab55f1ca71 ("* apt/dpkg: Recognize options in apt sources.list. (LP: #1238620)")

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -1402,8 +1402,11 @@ class __AptDpkgPackageInfo(PackageInfo):
                 fields = line.split()
                 if len(fields) >= 3 and fields[0] == "deb":
                     if fields[1].startswith("["):
-                        # options given, mirror is in third field
-                        mirror_idx = 2
+                        # mirror follows options
+                        for mirror_idx in range(1, len(fields) - 1):
+                            if fields[mirror_idx].endswith("]"):
+                                mirror_idx = mirror_idx + 1
+                                break
                     else:
                         mirror_idx = 1
                     if fields[mirror_idx].startswith("http://") or fields[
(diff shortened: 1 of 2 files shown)


bdrung:
Thanks for developing a fix. Please add a test case for the crash. Maybe the better fix is to offload the parsing to python-apt instead?

schopin-pro:
Using python-apt would give us deb822 compat for free, too :)

xypron:
The developers of python-apt don't document classes and methods, ouch.

xypron:
This finds the url:
```
import aptsources.sourceslist
source_entry = aptsources.sourceslist.SourceEntry('deb [ arch=riscv64 ] https://example.com/path release section')
print(source_entry.uri)
```

xypron:
> Using python-apt would give us deb822 compat for free, too :)

This would require changing _get_mirror() which uses hard coded "/etc/apt/sources.list". DEB822 sources have the extension .sources.

bdrung:
> This finds the url:
> 
> ```
> import aptsources.sourceslist
> source_entry = aptsources.sourceslist.SourceEntry('deb [ arch=riscv64 ] https://example.com/path release section')
> print(source_entry.uri)
> ```

This code snippet is what we want to use instead of extending our parser further. I used your research and your test case and developed a fix that uses `aptsources.sourceslist.SourceEntry`: https://github.com/canonical/apport/pull/154. Some of your test cases are not needed since they are already covered by `test_mirror_from_apt_sources`.

bdrung:
Superseded by https://github.com/canonical/apport/pull/154

PR #149, opened 2023-03-28
refactor: Introduce run_as_real_user()

bdrung:
Move calling `xdg-open` in `UserInterface.open_url` into a separate function `run_as_real_user`. Also simplify the call to `pgrep`: Call with `-n` to get only one process ID and drop `-a` to get only the process ID as output.


schopin-pro (approved):
LGTM

PR #150, opened 2023-03-30
feat: allow recent_syslog to collect all journal messages that the user can see

ogayot:
When using journalctl to extract the "recent" syslog messages, the `--system` option was unconditionally passed. This means messages from the user journal are not included. Added a parameter of `recent_syslog` to /not/ pass the `--system` option to journalctl ; so that everything the current user sees (including output of commands run with `systemd-run --user`) is included if requested.

To use the new option, the caller must invoke `recent_syslog` with the `journald_system_only=False` keyword argument. By default, its value is set to True so that the default behavior remains unchanged.

This is something we'd need for subiquity since we rely on `systemd-run --user` quite often.

### DIFF
--- a/apport/hookutils.py
+++ b/apport/hookutils.py
@@ -622,17 +622,23 @@ def __filter_re_process(pattern, process):
     return ""
 
 
-def recent_syslog(pattern, path=None):
+def recent_syslog(pattern, path=None, *, journald_only_system=True):
     """Extract recent system messages which match a regex.
 
     pattern should be a "re" object. By default, messages are read from
     the systemd journal, or /var/log/syslog; but when giving "path", messages
     are read from there instead.
+    The journald_only_system parameter controls the scope of messages that are
+    extracted when reading from the systemd journal. If set to True (the
+    default), only messages from the system services are extracted. If set to
+    False, all messages that the current user can see are extracted.
     """
     if path:
         command = ["tail", "-n", "10000", path]
     elif os.path.exists("/run/systemd/system"):
-        command = ["journalctl", "--system", "--quiet", "-b", "-a"]
+        command = ["journalctl", "--quiet", "-b", "-a"]
+        if journald_only_system:
+            command.append("--system")
     elif os.access("/var/log/syslog", os.R_OK):
         command = ["tail", "-n", "10000", "/var/log/syslog"]
     else:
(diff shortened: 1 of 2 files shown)


apport/hookutils.py
bdrung, line 1: Can you document the journald_only_system parameter in the docstring?

bdrung (changes requested):
praise: I am impressed by merge request: Well written commit message and test coverage! It does not happen often that the first iteration passes my critical review.

apport/hookutils.py
```diff
@@ -622,7 +622,7 @@ def __filter_re_process(pattern, process):
     return ""
 
 
-def recent_syslog(pattern, path=None):
+def recent_syslog(pattern, path=None, *, journald_only_system=True):
```
schopin-pro, line 625: question (not-blocking): why the wildcard here?

I'm not a Python expert, please educate me :)

schopin-pro (approved):
LGTM.

Just to clarify the use case here: subiquity adds a custom hook and would like to use this in it, correct?

apport/hookutils.py
bdrung: The wildcard means that the following parameters are keyword only arguments. So calling `recent_syslog(pattern, path, False)` is not allowed.

apport/hookutils.py
ogayot: Sure thing, documentation added!

apport/hookutils.py
ogayot: > LGTM.
> 
> Just to clarify the use case here: subiquity adds a custom hook and would like to use this in it, correct?

Yes, that's correct. We do have a hook in  `/usr/share/apport/package-hooks/subiquity.py` which is very much outdated. But we also use apport explicitly in subiquity and are currently trying to get the journal included in our bug reports. See  https://github.com/canonical/subiquity/pull/1625

bdrung (approved):
Thanks. Looks great now. I will merge once the CI runs are done.

ogayot:
> LGTM.
> 
> Just to clarify the use case here: subiquity adds a custom hook and would like to use this in it, correct?

Yes, that's correct. We do have a hook in  `/usr/share/apport/package-hooks/subiquity.py` which is very much outdated. But we also use apport explicitly in subiquity and are currently trying to get the journal included in our bug reports. See  https://github.com/canonical/subiquity/pull/1625

PR #151, opened 2023-03-30
fix(apport-kde): Fix showing progress in the progress bar

bdrung:
Commit 57c674a6d906a6e66a5c795aa45710cd2844d24d tries to fix following crash:

```
$ ubuntu-bug dolphin
  File "/usr/share/apport/apport-kde", line 532, in <module>
    sys.exit(UserInterface.run_argv())
  File "/usr/lib/python3/dist-packages/apport/ui.py", line 717, in run_argv
    return self.run_report_bug()
  File "/usr/lib/python3/dist-packages/apport/ui.py", line 558, in run_report_bug
    self.file_report()
  File "/usr/lib/python3/dist-packages/apport/ui.py", line 1391, in file_report
    self.ui_set_upload_progress(__upload_progress)
  File "/usr/share/apport/apport-kde", line 442, in ui_set_upload_progress
    self.progress.set(progress)
  File "/usr/share/apport/apport-kde", line 129, in set
    progress.setValue(value * 1000)
TypeError: setValue(self, int): argument 1 has unexpected type 'float'
```

The value is a float between 0 and 1. Converting the value to integer before multiplying by 1000 result only two values for the progress bar: 0% and 100%.

### DIFF
--- a/apport/ui.py
+++ b/apport/ui.py
@@ -2075,7 +2075,7 @@ class UserInterface:
             "this function must be overridden by subclasses"
         )
 
-    def ui_set_upload_progress(self, progress):
+    def ui_set_upload_progress(self, progress: typing.Optional[float]) -> None:
         """Update data upload progress bar.
 
         Set the progress bar in the debug data upload progress window to the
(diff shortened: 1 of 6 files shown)


LaurentBonnaud:
Thanks a lot for this fix!

schopin-pro (approved):

PR #152, opened 2023-03-30
fix: Only open browser as user (via sudo) if running as root

bdrung:
By setting the environment variable `SUDO_UID` or `PKEXEC_UID` Apport can be tricked into calling `xdg-open` via sudo for the specified user ID.

Prevent that unexpected behavior if Apport was not called as root.


schopin-pro (approved):
LGTM

PR #153, opened 2023-04-04
Replace sudo by dropping privileges ourselves

bdrung:
In 2012 commit c178f28c505f1249f525d6441bd21469c810b14b reverted back to calling sudo instead of dropping privileges ourselves due to Launchpad bug #916810.

This behavior of Firefox cannot be reproduced any more. Stop relying on sudo and drop the privileges ourselves. This reduces the dependency on `sudo` and relying on `sudo` to be properly configured.

### DIFF
--- a/apport/ui.py
+++ b/apport/ui.py
@@ -24,6 +24,7 @@ import gzip
 import io
 import locale
 import os.path
+import pwd
 import queue
 import re
 import shutil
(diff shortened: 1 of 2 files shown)


schopin-pro (commented):
The code LGTM, but I'm a bit confused as to *why* we want this?

bdrung:
We reduce the dependency on  `sudo` and do not rely on `sudo` to be properly configured.

PR #159, opened 2023-04-12
chore: Add todos for functions/methods that should be split

bdrung:
To prevent introducing big functions, add `TODO`s for functions/methods that should be split and remove the override from the global pylint configuration.


schopin-pro (approved):
I like this!

PR #164, opened 2023-04-19
refactor(apport): Move get_core_path out of write_user_coredump

bdrung:
To reduce the amount of needed parameters, move the `apport.fileutils.get_core_path` call out of `write_user_coredump`. Log the full core path in `write_user_coredump` to avoid needing `core_name`.


data/apport
```diff
@@ -255,7 +256,7 @@ def write_user_coredump(
     except OSError:
         return
 
-    logger.error("writing core dump to %s (limit: %s)", core_name, str(limit))
+    logger.error("writing core dump to %s (limit: %s)", core_path, str(limit))
```
schopin-pro, line 259: thought: this probably shouldn't be an error though, should it?

(irrelevant for the current PR)

schopin-pro (approved):
LGTM.

We really should do a pass on the log levels at some point, it seems weird that everything is logged at error level! Not the subject at hand though, so merging.

data/apport
bdrung: Here you go: https://github.com/canonical/apport/pull/169

PR #169, opened 2023-04-20
fix(apport): Demote non-error log messages to info level

bdrung:
Apport only used error log message even for purely informational logs. Since `logging` is used nowadays, demote the non-error log messages to info level. Reduce the log level from warning to info to still print these messages.


schopin-pro (approved):
YES! That sounds much more reasonable

PR #171, opened 2023-04-21
fix(tests): Fix check for full name in test_run_crash_anonymity

bdrung:
The value `pw_gecos` contains a comma separated list where the first element might be the full user name. Split the first element from `pw_gecos` (if it is not empty) to check that the full name is not in the crash report.


schopin-pro (approved):
LGTM.

Methinks you went a bit overboard with the fine commits though. The first one could have been folded into the `_get_sensitive_strings()` introductory one ;)

PR #174, opened 2023-04-27
fix(tests): Fix false positives for test_run_crash_anonymity

bdrung:
The test `test_run_crash_anonymity` can fail if the user or host name is short:

```
======================================================================
FAIL: test_run_crash_anonymity (tests.integration.test_ui.T)
run_crash() anonymization
----------------------------------------------------------------------
  File "/usr/lib/python3.10/unittest/mock.py", line 1369, in patched
    return func(*newargs, **newkeywargs)
  File "/<<PKGBUILDDIR>>/tests/integration/test_ui.py", line 1650, in test_run_crash_anonymity
    self.assertNotIn(
AssertionError: 'kin' unexpectedly found in 'ProblemType: Crash\nArchitecture: amd64\n[...]'
```

The string `kin` can be found in the report (for example in the tag `kinetic`). So only search for full words in the report if the string is short (less than five characters and not the string `ubuntu`).

### DIFF
--- a/tests/integration/test_ui.py
+++ b/tests/integration/test_ui.py
@@ -9,6 +9,7 @@ import io
 import locale
 import os
 import pwd
+import re
 import shutil
 import signal
 import stat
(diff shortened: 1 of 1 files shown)


schopin-pro (approved):

schopin-pro:
Wait, on second thought: what's the username in our CI runs?

bdrung:
From the home directory name in `action/checkout@v3` I assume the username is `runner`.

PR #187, opened 2023-05-16
small encapsulation change

BMaxV:
Hi, 

this is a small change to say hello and test the water, learn the process.

The function I'm doing it to even has a "# TODO: Split into smaller functions/methods". So we'll see.

### DIFF
--- a/problem_report.py
+++ b/problem_report.py
@@ -391,36 +391,8 @@ class ProblemReport(collections.UserDict):
         Files are written in RFC822 format.
         """
         self._assert_bin_mode(file)
-
-        # sort keys into ASCII non-ASCII/binary attachment ones, so that
-        # the base64 ones appear last in the report
-        asckeys = []
-        binkeys = []
-        for k in self.data.keys():
-            if only_new and k in self.old_keys:
-                continue
-            v = self.data[k]
-            if hasattr(v, "find"):
-                if self.is_binary(v):
-                    binkeys.append(k)
-                else:
-                    asckeys.append(k)
-            else:
-                if (
-                    not isinstance(v, CompressedValue)
-                    and len(v) >= 2
-                    and not v[1]
-                ):
-                    # force uncompressed
-                    asckeys.append(k)
-                else:
-                    binkeys.append(k)
-
-        asckeys.sort()
-        if "ProblemType" in asckeys:
-            asckeys.remove("ProblemType")
-            asckeys.insert(0, "ProblemType")
-        binkeys.sort()
+        
+        asckeys, binkeys = self.sort_keys()
 
         # write the ASCII keys first
         for k in asckeys:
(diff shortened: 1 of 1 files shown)


problem_report.py
```diff
@@ -550,6 +522,40 @@ def write(self, file, only_new=False):
                 file.write(base64.b64encode(block))
                 file.write(b"\n")
 
+    def sort_keys(self, only_new):
```
bdrung, line 525: suggestion: Can you add type hints to the new function? I am doing that for new code to apport.

```suggestion
    def sort_keys(self, only_new: bool) -> typing.Tuple[list[str], list[str]]:
```

bdrung, line 525: question: The name `sort_keys` mislead me to think that this function changes the object, but it only a getter function. Maybe name it `get_sorted_keys`?

bdrung, line 525: issue: Is it intentional that you made this method public? If not, please make it private by prepending an underscore the name. This allows us to refactor this method in the future without breaking consumers of the library.

bdrung (changes requested):
I'll try to use https://conventionalcomments.org/ for the review.

praise: Thanks for your contribution. It's great to see external contribution.

polish: The commit message "small encapsulation change to say hello and test the water" shows your intention but is not very useful in the future when looking at the git history. I recommend to follow https://www.conventionalcommits.org and prefix the commit with `refactor`. Comparing with previous commit I would have named them `refactor: Introduce ProblemReport.sort_keys` or `refactor: Split sort_keys from ProblemReport.write`

problem_report.py
BMaxV: I can do that.

I dislike them, as I think they are misleading, because the type checks that can be done don't guarantee anything at execution time.

But if that is what it takes, I can do that.

BMaxV:
The lint tests are failing for some reason I don't get and it's not obvious how to do the linting tests locally? From the reports I could figure out how to run the regular tests locally.

problem_report.py
bdrung: I see two benefits of them: Added documentation for the caller and finding some (but not all) type issues. We will run mypy on the code find them (see https://github.com/canonical/apport/pull/184). We found 2-3 real bugs thanks to mypy.

BMaxV:
Regarding the linting installing `pydocstyle` wasn't a problem. So I test and fix that. But `pyflakes` isn't recognized as `pyflakes3`. Not sure what to do about that. There is an apt package but that would be mixing pip and apt.

Let me know if there are more kinds of things to do or if it's just that pydocstyle lint thing.

bdrung:
We use the Debian packages for the linters. If you prefer to use pip only, you could modify the shell script to look for `pyflakes` if `pyflakes3` cannot be found.

PR #188, opened 2023-05-18
refactor: add _write_asc_to_file

BMaxV:
Hello,

here is another one. Should be straight forward. 

Something I noticed when looking at this, was that `fileutils` constructs the path name and opens the file. And then the file handle gets passed around. That surprised me a bit , I would have expected something like:

```
my_data = some_function()
my_path = path_creation_function()
with open(my_path) as f:
    f.write(my_data)
```

Should that be a todo for the future?

### DIFF
--- a/problem_report.py
+++ b/problem_report.py
@@ -394,45 +394,8 @@ class ProblemReport(collections.UserDict):
 
         asckeys, binkeys = self._get_sorted_keys(only_new)
 
-        # write the ASCII keys first
         for k in asckeys:
-            v = self.data[k]
-
-            # if it's a tuple, we have a file reference; read the contents
-            if not hasattr(v, "find"):
-                if len(v) >= 3 and v[2] is not None:
-                    limit = v[2]
-                else:
-                    limit = None
-
-                fail_on_empty = len(v) >= 4 and v[3]
-
-                if hasattr(v[0], "read"):
-                    v = v[0].read()  # file-like object
-                else:
-                    with open(v[0], "rb") as f:  # file name
-                        v = f.read()
-
-                if fail_on_empty and len(v) == 0:
-                    raise OSError("did not get any data for field " + k)
-
-                if limit is not None and len(v) > limit:
-                    del self.data[k]
-                    continue
-
-            if isinstance(v, str):
-                # unicode → str
-                v = v.encode("UTF-8")
-
-            file.write(k.encode("ASCII"))
-            if b"\n" in v:
-                # multiline value
-                file.write(b":\n ")
-                file.write(v.replace(b"\n", b"\n "))
-            else:
-                file.write(b": ")
-                file.write(v)
-            file.write(b"\n")
+            self._write_key_and_ascii_value_to_file(file, k)
 
         # now write the binary keys with gzip compression and base64 encoding
         for k in binkeys:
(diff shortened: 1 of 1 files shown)


problem_report.py
```diff
@@ -557,6 +518,50 @@ def _get_sorted_keys(
         binkeys.sort()
         return asckeys, binkeys
 
+    def _write_asckeys_to_file(
```
bdrung, line 521: suggestion: `asckeys` is a little cryptic. How about using the long form `ascii_keys`?

problem_report.py
```diff
@@ -557,6 +518,50 @@ def _get_sorted_keys(
...
+        self, file: typing.BinaryIO, asckeys: list[str]
+    ):
+        # write the ASCII keys first
```
bdrung, line 524: suggestion: Shouldn't this comment stay in the calling function?

problem_report.py
```diff
@@ -557,6 +518,50 @@ def _get_sorted_keys(
...
+
+        for k in asckeys:
```
bdrung, line 526: nitpick/suggestion: To reduce the level of indentation in this method, the for loop could be moved out of this method (by either keeping the loop in the calling function or add another function just for looping over it).

bdrung (changes requested):
Thanks for your merge request.

The Apport project is covered by Canonical’s contributor agreement (see https://canonical.com/projects/directory). Before I can merge the change, you need to sign the [Canonical’s contributor agreement](https://www.ubuntu.com/legal/contributors).

TODO for myself: Add a contribution document stating that the contributor agreement needs to be signed and check it with the CI.

bdrung:
Forgot to mention: Open the file in one function and requiring the caller to close it is a bad pattern. So improving that is welcome (but beware to not break the API).

problem_report.py
BMaxV: I can put it there.

In the end the function should look something like:

```
def write(self, file, only_new=False):
    self._get_sorted_keys(only_new)()
    self._write_asci()
    self._write_binary()
    self._rest()
```

_write_binary() https://github.com/canonical/apport/blob/main/problem_report.py#L437
_rest() https://github.com/canonical/apport/blob/main/problem_report.py#L514

write binary might have it's subfunctions too. 

It should be really obvious that this is happening first and ASCII is right in the name. 

This:

```
def write(self, file, only_new=False):
    self._get_sorted_keys(only_new)()
    # write ascii first
    self._write_asci()
    # write binary second
    self._write_binary()
    self._rest()
```

 wouldn't need the comment.

bdrung: suggestion: Yes, the final write method will be small enough to understand it without that comment. Adding a comment there will reduce the readability. So I suggest to drop that comment completely.

BMaxV:
I sent the contributor agreement as well.

problem_report.py
```diff
@@ -557,6 +520,47 @@ def _get_sorted_keys(
         binkeys.sort()
         return asckeys, binkeys
 
+    def _write_ascii_key_to_file(
```
bdrung, line 523: nitpick(name): The name just says that it will write the key, but it writes the key and the ASCII value. So `_write_key_and_ascii_value_to_file` would decribe it more precisely.

bdrung (approved):
Thanks. Last request before merging: Can you squash the two commits since we are using fast-forward merges?

BMaxV:
I did both. :)

bdrung:
Thanks, but the git commit message plus message body does not match the diff any more (renamed method). Can you rewrite the git commit message to reflect the current name?

PR #189, opened 2023-05-22
refactor: add _make_compressed_block

BMaxV:
This one is again relatively simple.

It's a bit different since the function doesn't do *exactly* the same thing, I'm keeping the `file.write`s in the main function. Instead the function makes the binary string(?) that's being written.

I spent like 3 hours reading `.load` and `.write`.

https://github.com/canonical/apport/blob/main/problem_report.py#L411

that is basically an `if` and a very big `else`. I don't think I can make a helpful or meaningful change without rewriting that part quite a bit.

Let me know if that's something you want to review or if I should keep making simple changes.

### DIFF
--- a/problem_report.py
+++ b/problem_report.py
@@ -368,6 +368,93 @@ class ProblemReport(collections.UserDict):
                 return value
         return value
 
+    def _case_complex_value(self, k, v):
+        limit = None
+        size = 0
+
+        mainblock = k.encode("ASCII")
+        mainblock += b": base64\n "
+
+        # write gzip header
+        gzip_header = (
+            GZIP_HEADER_START
+            + b"\010\000\000\000\000\002\377"
+            + k.encode("UTF-8")
+            + b"\000"
+        )
+        mainblock += base64.b64encode(gzip_header)
+        mainblock += b"\n "
+        crc = zlib.crc32(b"")
+
+        bc = zlib.compressobj(6, wbits=-zlib.MAX_WBITS)
+        
+        # direct value
+        if hasattr(v, "find"):
+            size += len(v)
+            crc = zlib.crc32(v, crc)
+
+            compressblock = bc.compress(v)
+            if compressblock:
+                mainblock += base64.b64encode(compressblock)
+                mainblock += b"\n "
+
+        # file reference
+        else:
+            if len(v) >= 3 and v[2] is not None:
+                limit = v[2]
+
+            if hasattr(v[0], "read"):
+                f = v[0]  # file-like object
+            else:
+                # hard to change, pylint: disable=consider-using-with
+                f = open(v[0], "rb")  # file name
+            while True:
+                block = f.read(1048576)
+                size += len(block)
+                crc = zlib.crc32(block, crc)
+                if limit is not None:
+                    if size > limit:
+                        # roll back
+                        # don't need to roll back, because nothing was written
+                        return None
+                if block:
+                    compressblock = bc.compress(block)
+                    if compressblock:
+                        mainblock += base64.b64encode(compressblock)
+                        mainblock += b"\n "
+                else:
+                    break
+            if not hasattr(v[0], "read"):
+                f.close()
+
+            if len(v) >= 4 and v[3]:
+                if size == 0:
+                    raise OSError(
+                        "did not get any data for field %s from %s"
+                        % (k, str(v[0]))
+                    )
+
+        # flush compressor and write the rest
+        if not limit or size <= limit:
+            block = bc.flush()
+            # append gzip trailer: crc (32 bit) and size (32 bit)
+            if crc:
+                block += struct.pack("<L", crc & 0xFFFFFFFF)
+                block += struct.pack("<L", size & 0xFFFFFFFF)
+
+            mainblock += base64.b64encode(block)
+            mainblock += b"\n"
+
+        return mainblock
+
+    @staticmethod
+    def _case_compressed_value(k, v):
+        block = k.encode("ASCII")
+        block += b": base64\n "
+        block += base64.b64encode(v.gzipvalue)
+        block += b"\n"
+        return block
+
     def write(self, file, only_new=False):
         # TODO: Split into smaller functions/methods
         # pylint: disable=too-many-branches,too-many-locals,too-many-statements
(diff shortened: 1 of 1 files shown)


problem_report.py
```diff
@@ -485,6 +479,28 @@ def write(self, file, only_new=False):
...
+        self, v: typing.BinaryIO, bc
+    ) -> typing.BinaryIO:
+        block = b""
+        outblock = bc.compress(v)
+        if outblock:
+            block += base64.b64encode(outblock)
+            block += b"\n "
+        return block
+
+    def _make_gzip_header_block(self, k: typing.BinaryIO) -> typing.BinaryIO:
```
bdrung, line 492: question: Isn't the key `k` of type `str` and the return value just `bytes`?

problem_report.py
```diff
@@ -485,6 +479,28 @@ def write(self, file, only_new=False):
...
+        block = b""
```
bdrung, line 493: question: Why defining an empty bytes string here? Can't it be assigned below?

problem_report.py
bdrung, line 492: issue: The linter complains about not used `self`. The method can be annotated with `@staticmethod` to address it.

problem_report.py
```diff
@@ -485,6 +479,28 @@ def write(self, file, only_new=False):
...
 
+    def _make_compressed_block(
+        self, v: typing.BinaryIO, bc
+    ) -> typing.BinaryIO:
+        block = b""
+        outblock = bc.compress(v)
+        if outblock:
+            block += base64.b64encode(outblock)
+            block += b"\n "
+        return block
```
bdrung, line 490: nitpick: Inverting the if-condition would make the method use the return early pattern that makes the code a bit more readable:
```suggestion
        if not outblock:
            return b""
        block = base64.b64encode(outblock)
        block += b"\n "
        return block
```
I found this blog post explaining the reasoning quite well: https://medium.com/swlh/return-early-pattern-3d18a41bba8

bdrung (changes requested):
Thanks for working on improving the code. I remember that I also spend quite some time on that part to understand it. So making it easier to understand will help the next person looking at it.

praise: Carving out those two helper methods look good to me and that they operate on byte strings.

The commit title `refactor: add _make_gzip_header_block _make_compressed_block` mislead me on the first glance since it states two functions without `and`. Using `and` in a commit often indicates that two or more things are changed in one commit that could be split into multiple commits. The two methods are not coupled. Could you split them into two separate commits / merge requests?

BMaxV:
As suggested, I rewrote the commit to only change one function.

problem_report.py
```diff
@@ -485,6 +479,16 @@ def write(self, file, only_new=False):
                 file.write(base64.b64encode(block))
                 file.write(b"\n")
 
+    @staticmethod
+    def _make_compressed_block(v: bytes, bc: zlib.compressobj) -> bytes:
```
bdrung, line 483: issue: Mypy complains about it:
```
problem_report.py:483: error: Function "zlib.compressobj" is not valid as a type  [valid-type]
problem_report.py:483: note: Perhaps you need "Callable[...]" or a callback protocol?
problem_report.py:485: error: zlib.compressobj? has no attribute "compress"  [attr-defined]
Found 2 errors in 1 file (checked 1 source file)
```
I am not sure how to address them. `type(bc)` says `zlib.Compress` but mypy says that this type is not defined. If not possible, the annotation for `bc` needs to be removed.

idea: Other option would be to encapsulate the compressor in a separate class. That class could contain the crc and compression. Then that class could have a "write header", "write block", "write the rest", "roll back", etc. The result could be more readable, but it could be opposite as well.

problem_report.py
```diff
@@ -485,6 +479,16 @@ def write(self, file, only_new=False):
...
+        block = b""
+        outblock = bc.compress(v)
+        if not outblock:
+            return block
+        block += base64.b64encode(outblock)
```
bdrung, line 488: suggestion: The first `block = b""` assignment can be avoided (making the code shorter):
```suggestion
        outblock = bc.compress(v)
        if not outblock:
            return b""
        block = base64.b64encode(outblock)
```

problem_report.py
bdrung, line 483: idea: `_make_compressed_block` only tells part of what the function does. It also base64-encodes the block and also adds a newline at the end. So something in the way of "make compressed base64 encoded line" would describe it more precisely.

bdrung (changes requested):
The change looks good, except that mypy complains.

problem_report.py
BMaxV: Maybe passing the compression object anywhere is the wrong approach after all.

Let me rewrite the whole thing as a suggestion, it's probably easier to explain my idea about it that way.

BMaxV:
Ok. Here is another approach.

And I know it probably doesn't pass the requirements, this is just to explain what I want to do.

There are three cases:

(1)  it's an easy compressed value, can be encoded put into bytes and that's it.

(2) it's below [this comment](https://github.com/canonical/apport/blob/main/problem_report.py#L430) in line 430, it is not compressed yet, so it gets compressed, at the end the compressor is flushed and the value, the size and the checksum are written.

(3) [this else](https://github.com/canonical/apport/blob/main/problem_report.py#L439) in line 439, where v[0] is either file-like or a file. Which can lead to `k` and `v` being discarded.

The last commit separates (1) into a function that returns bytes.

And it separates (2) and (3) into a function that returns either bytes or None, which is (3)'s discard case.

The big benefits here are:
 * writing to the file only happens in one place, it's no longer necessary to truncate, because all writing happens after the limit check.
 * there is no need to worry about the type of zlib.compress or crc, because those are not passed anywhere.

(the tests pass on my machine so I don't know what's up with that)

bdrung:
issue: The core dump can become quite big (several gigabytes). So the compressed blocks should be written into the file and not kept in memory. Otherwise apport could run out of memory.

Some of the test failed to upload the coverage report. You could just retry them.

BMaxV:
Ok, that's interesting. Let me see how I can solve that...

BMaxV:
So, the solution is I don't touch that for now, instead I separated the cases as I described above. 

The next steps can be putting those cases into their own functions.

I didn't add any new functions.  I moved variables closer to where they are used and removed them when they are not needed for that case.

 * size in case 2 is just the length of v, no need to initialize at 0
 * case 2 probably doesn't need a zlib.compress**obj**, because it's only compressing once

I changed the inputs to compressobj because most arguments are just the default values. [link to the docs](https://docs.python.org/3/library/zlib.html#zlib.compress)

 * limit is only needed in case 3
 *  limit in case 3 is either v[2] or None
 * `curr_pos` is only needed in case 3

There are some parts that are now duplicate code, but for now it's beneficial that there are three clearly separate cases we can look at in isolation.

bdrung:
issue: `limit` will be `None` for case 2. So the compressor flush will be used there.

This refactoring is hard to review since the refactoring makes the code worse and only following commits will make it better again. For the review it would be easier to start carving out the functions step by step instead. Do you have the final result somewhere to see where this is going?

bdrung:
For retying the test run: I can see a "Re-run jobs" button on the top right on https://github.com/canonical/apport/actions/runs/5058074551. Can you see that as well or does that require special permissions?

bdrung:
> In the function as it is now in `origin main`, it's defined in line [403](https://github.com/canonical/apport/blob/main/problem_report.py#L403), set to `None` and then never changes for that case. `if not limit or size <= limit:` in line [478](https://github.com/canonical/apport/blob/main/problem_report.py#L478) is always `True`. So the compressor flush is always executed. I don't know if that's correct or intended.

Yes. That is intended. The flush should always be done. The only exception: The size exceeds the limit and the write will be rolled back. So refactoring the code to `return` in the rollback condition would make it more readable. Then the condition for the flush can be dropped and the flush be done always.

nitpick: My gut feeling does not like this refactoring and it's hard to exactly tell why. Probably because the code is just split into multiple functions but not made easier to read. Method names like `_write_case_file_reference` does not tell me what that method does. If finding good descriptive names for the function is an indicator that the function should be refactored.

suggestion: Maybe we end up with a better structure if we try to address the code duplication first and carve out smaller methods (like you did initially). For example I see a lot of
```
file.write(base64.b64encode(block))
file.write(b"\n ")
```
That could be put into a `_write_b64encoded_line`. And instead of writing `\n` with a following space, we could write full lines starting with the space. Then `_write_b64encoded_line(gzip_header)` could be used for the gzip header and `_write_b64encoded_line(v.gzipvalue) for case (1).

Writing
```python
block = k.encode("ASCII")
block += b": base64\n "
```
is needed in all cases. So better do it only once instead of duplicating it.

BMaxV:
Gut feelings are good.

I agree that the case (3) function  `_write_case_file_reference` is still awkward to read.

My biggest "felt open question" while reading is the loop:

It is not obvious to me when limit exists and how big it is.

The "sometimes it's a file and sometimes it isn't" that's solved with hasattr is strange, as well as the closing.

The reason I didn't put that into it's own function is that passing the compression with **typehints** is a challenge.

Let me do that real quick without typehints.

BMaxV:
As a side note, doing this makes it obvious that the check for size:

` if not limit or **size <= limit**`

is always True. So it could be removed. The other case of `size>limit` is checked inside the loop function and if that's the case, the file is truncated and both functions exist.

bdrung:
> suggestion: Maybe we end up with a better structure if we try to address the code duplication first and carve out smaller methods (like you did initially). For example I see a lot of
> 
> ```python
> file.write(base64.b64encode(block))
> file.write(b"\n ")
> ```
> 
> That could be put into a `_write_b64encoded_line`. And instead of writing `\n` with a following space, we could write full lines starting with the space. Then `_write_b64encoded_line(gzip_header)` could be used for the gzip header and `_write_b64encoded_line(v.gzipvalue) for case (1).

I implemented that idea in https://github.com/canonical/apport/pull/238

schopin-pro:
Hi!

Thank you for your contributions. However, in this case, neither Benjamin nor I see the resulting code as a net improvement, and this it has been inactive for some months now, I'll just close it out.

Again, thank you for spending time working on this. If anything, this entire PR is actually an unexpected but fairly decent documentation of the existing code, which is always nice to have, even in as odd a place.

(also, just because I close it out doesn't mean we can't reopen this or use a fresh PR if you want to pick it up again)

PR #190, opened 2023-06-16
fix: Fix spelling mistakes found by codespell

bdrung:
`codespell .` finds several spelling mistakes. Address all of them (except for two false positives that can be addressed later).


schopin-pro (approved):

bdrung:
I suggest to ignore codecov/patch for this one.

PR #193, opened 2023-07-06
Implement APT deb822 support (FR-4091)

schopin-pro:
There's a (somewhat) new format for sources definition supported by APT, based on the deb822 syntax that's widely used in the Debian world. The APT developpers would like to make that new format the default (there are plenty of good reasons), and for that we need to support it.

Luckily some of the heavy lifting has already been done in `python-apt`.

Note that in order to easily test the new format, I migrated `tests/system/test_packaging_apt_dpkg.py` from using the unittest API to directly use pytest's API, since we're already using it as test runner anyway (split off in #194). This allowed me to simply parameterize the tests over the source format, ensuring a good test coverage. Said migration should account for a good chunk of the test file diff.

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -16,6 +16,12 @@ This is used on Debian and derivatives such as Ubuntu.
 # TODO: Address following pylint complaints
 # pylint: disable=invalid-name
 
+# From Python 3.12 on, it doesn't try to evaluate type signatures at runtime
+# anymore. This behaviour is helpful to us as apt_sl.Deb822SourceEntry isn't
+# always a valid type, so use the __future__ mechanism to have this behaviour
+# in older versions as well.
+from __future__ import annotations
+
 import contextlib
 import datetime
 import glob
@@ -39,13 +45,42 @@ import urllib.request
 from typing import Optional
 
 import apt
+import aptsources.sourceslist as apt_sl
 from aptsources.sourceslist import SourceEntry
 
 import apport.logging
 from apport.packaging import PackageInfo
 
+WITH_DEB822_SUPPORT = hasattr(apt_sl, "Deb822SourceEntry")
 
-def _load_all_sources(apt_dir: str) -> list[SourceEntry]:
+
+# pylint: disable-next=no-member
+def _parse_deb822_sources(source: str) -> list[apt_sl.Deb822SourceEntry]:
+    if not WITH_DEB822_SUPPORT:
+        return []
+
+    sections = []
+    current = []
+    with open(source, encoding="utf-8") as f:
+        for line in f.read().split("\n"):
+            line = line.rstrip()
+            if line:
+                current.append(line)
+            else:
+                if not current:
+                    continue
+                # pylint: disable-next=no-member
+                sections.append(apt_sl.Deb822SourceEntry("\n".join(current), source))
+                current = []
+
+        if current:
+            # pylint: disable-next=no-member
+            sections.append(apt_sl.Deb822SourceEntry("\n".join(current), source))
+        return sections
+
+
+# pylint: disable-next=no-member
+def _load_all_sources(apt_dir: str) -> list[apt_sl.Deb822SourceEntry | SourceEntry]:
     """Given an APT configuration directory (e.g. /etc/apt/), loads up all
     the source data in one easy-to-consume list.
     """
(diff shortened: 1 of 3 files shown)


bdrung:
It would be useful to specify the minimum required python-apt version in the README.

schopin-pro:
Re: specifying dependency info in the README, there's no mention of any dep in it so far. We're not even mentioning the minimum version of Python we're supporting!

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -955,15 +1093,16 @@ def _copy_ubuntu_keyrings(keyring_dir: str) -> None:
 
 
 def _write_source_file(
-    sources_filename: str, uri: str, release: str, updates: bool
+    sources_filename: str, uri: str, release: str, updates: bool, ppa=False
```
bdrung, line 1096:
```suggestion
    sources_filename: str, uri: str, release: str, updates: bool, ppa: bool = False
```

apport/packaging_impl/apt_dpkg.py
```diff
@@ -1607,24 +1649,46 @@ def create_ppa_source_from_origin(
...
+            try:
+                with contextlib.closing(urllib.request.urlopen(debug_url)) as response:
+                    response.read()
+                debug_entry = apt_sl.Deb822SourceEntry(None, "")
+                debug_entry.uris = [
+                    f"http://ppa.launchpad.net/{user}/{ppa_name}/{distro}"
+                ]
+                debug_entry.types = ["deb"]
+                debug_entry.comps = ["main/debug"]
+                debug_entry.suites = [release_codename]
```
bdrung, line 1673: question: Why not just `main_entry.comps.append("main/debug")`?

schopin-pro: Because the main entry has deb-src, which isn't enabled on main/debug components.

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -701,19 +717,41 @@ def test_create_sources_for_a_named_ppa(configdir, rootdir, apt_style):
...
-    ) as f:
-        sources = f.read().splitlines()
-    assert (
-        "deb http://ppa.launchpad.net/daisy-pluckers/daisy-seeds/ubuntu"
-        " jammy main main/debug" in sources
-    )
-    assert (
-        "deb-src http://ppa.launchpad.net"
-        "/daisy-pluckers/daisy-seeds/ubuntu jammy main" in sources
-    )
```
bdrung, line 716: question: Is it intentional that you drop the deb-src assertion?

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -751,19 +789,41 @@ def test_create_sources_for_an_unnamed_ppa(configdir, rootdir, apt_style):
...
-    ) as f:
-        sources = f.read().splitlines()
-    assert (
-        "deb http://ppa.launchpad.net"
-        "/apport-hackers/apport-autopkgtests/ubuntu jammy main main/debug" in sources
-    )
-    assert (
-        "deb-src http://ppa.launchpad.net"
-        "/apport-hackers/apport-autopkgtests/ubuntu jammy main" in sources
-    )
```
bdrung, line 766: question: Is it intentional that you drop the deb-src assertion?

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -751,19 +789,41 @@ def test_create_sources_for_an_unnamed_ppa(configdir, rootdir, apt_style):
...
+    if WITH_DEB822_SUPPORT:
+        ppasource = os.path.join(
+            rootdir, "etc", "apt", "sources.list.d", f"{ppa}.sources"
+        )
+        entries = _parse_deb822_sources(ppasource)
+        assert next(
+            filter(
+                lambda e: set(["deb", "deb-src"]) == set(e.types)
```
bdrung, line 799: set comprehension:
```suggestion
                lambda e: {"deb", "deb-src"} == set(e.types)
```

tests/system/test_packaging_apt_dpkg.py
schopin-pro: No it's not.

schopin-pro: Fixed.

PR #194, opened 2023-07-06
Migrate the APT tests to pytest

schopin-pro:
Using idiomatic pytest is more concise than Python's builtin unittest framework, and will allow us to use some of pytest goodies, such as test parametrization, which is used to great effect in #193 

I also updated the testing README to remove any mention of `unittest`, as I expect this PR will shine the way to a new, unittest-free world ;)

### DIFF
--- a/tests/system/test_packaging_apt_dpkg.py
+++ b/tests/system/test_packaging_apt_dpkg.py
@@ -8,1056 +8,993 @@ import os
 import shutil
 import subprocess
 import tempfile
-import unittest
 
+import pytest
 from apt import apt_pkg
 
 from apport.packaging_impl.apt_dpkg import impl
 from tests.helper import has_internet, skip_if_command_is_missing
 
+if shutil.which("dpkg") is None:
+    pytest.skip("dpkg not installed", allow_module_level=True)
 
-@skip_if_command_is_missing("dpkg")
-class T(unittest.TestCase):
+
+@pytest.fixture(name="workdir")
+def fixture_workdir():
+    workdir = tempfile.mkdtemp()
+    yield workdir
+    shutil.rmtree(workdir)
+
+
+@pytest.fixture(name="cachedir")
+def fixture_cachedir(workdir):
+    ret = os.path.join(workdir, "cache")
+    os.makedirs(ret)
+    return ret
+
+
+@pytest.fixture(name="rootdir")
+def fixture_rootdir(workdir):
+    ret = os.path.join(workdir, "root")
+    os.makedirs(ret)
+    return ret
+
+
+@pytest.fixture(name="configdir")
+def fixture_configdir(workdir):
+    ret = os.path.join(workdir, "config")
+    os.makedirs(ret)
+    return ret
+
+
+@pytest.fixture(autouse=True)
+def environment(workdir):
+    orig_environ = os.environ.copy()
+    os.environ["HOME"] = workdir
+    yield
+    os.environ.clear()
+    os.environ.update(orig_environ)
+
+
+@pytest.fixture(autouse=True)
+def reset_impl():
     # pylint: disable=protected-access
+    orig_conf = impl.configuration
+    impl._apt_cache = None
+    impl._sandbox_apt_cache = None
+    yield
+    impl.configuration = orig_conf
 
-    def setUp(self):
-        # save and restore configuration file
-        self.orig_conf = impl.configuration
-        self.orig_environ = os.environ.copy()
-        self.workdir = tempfile.mkdtemp()
-        os.environ["HOME"] = self.workdir
-        self.cachedir = os.path.join(self.workdir, "cache")
-        self.rootdir = os.path.join(self.workdir, "root")
-        self.configdir = os.path.join(self.workdir, "config")
-        # reset internal caches between tests
-        impl._apt_cache = None
-        impl._sandbox_apt_cache = None
 
-    def tearDown(self):
-        impl.configuration = self.orig_conf
-        os.environ.clear()
-        os.environ.update(self.orig_environ)
-        shutil.rmtree(self.workdir)
+@pytest.mark.skipif(not has_internet(), reason="online test")
+def test_install_packages_versioned(configdir, cachedir, rootdir):
+    # TODO: Split into smaller functions/methods
+    # pylint: disable=too-many-locals,too-many-statements
+    """install_packages() with versions and with cache"""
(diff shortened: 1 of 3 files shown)


tests/system/test_packaging_apt_dpkg.py
```diff
@@ -444,23 +419,10 @@ def test_install_packages_permanent_sandbox(self):
...
-        )
-        self.assertEqual(
-            zonetab_written,
-            os.path.getctime(zonetab),
-            "zonetab written twice.",
-        )
-        self.assertTrue(
-            os.path.exists(os.path.join(self.rootdir, "usr/bin/stat"))
-        )
+        assert glob.glob(os.path.join(archives, "coreutils*.deb"))
```
bdrung, line 422: suggestion: Maybe add a comment above "# Check if the deb was downloaded" to keep the previous assertion string.

bdrung (approved):

tests/system/test_packaging_apt_dpkg.py
schopin-pro: Done :)

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -2,1062 +2,993 @@
...
+)
+
+
+@pytest.fixture()
+def fs():
+    # pylint: disable=attribute-defined-outside-init
+    # save and restore configuration file
+    workdir = tempfile.mkdtemp()
+    os.environ["HOME"] = workdir
+    fs = Fs(
```
bdrung, line 33: suggestion: `orig_conf` and `orig_environ` do not need to be public. They can be private to this function.

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -2,1062 +2,993 @@
...
+        os.environ.copy(),
+        workdir,
+        os.path.join(workdir, "cache"),
+        os.path.join(workdir, "root"),
+        os.path.join(workdir, "config"),
+    )
+    # reset internal caches between tests
+    impl._apt_cache = None
+    impl._sandbox_apt_cache = None
+    yield fs
```
bdrung, line 44: praise: The yield is much better than the setUp and tearDown functions.

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -2,1062 +2,993 @@
...
+if shutil.which("dpkg") is None:
+    pytest.skip(f"dpkg not installed", allow_module_level=True)
+
+Fs = collections.namedtuple(
+    "Fs", ["orig_conf", "orig_environ", "workdir", "cachedir", "rootdir", "configdir"]
+)
+
+
+@pytest.fixture()
+def fs():
```
bdrung, line 28: thought: This function could yield just the workdir and separate functions could return the other directories.

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -2,1062 +2,993 @@
...
+    # pylint: disable=attribute-defined-outside-init
```
bdrung, line 29: todo: That pylint override can be dropped.

bdrung (changes requested):

tests/system/test_packaging_apt_dpkg.py
schopin-pro: I disagree. Having all those util filesystem data computed at once and put into a pure data object is amply sufficient here. This is test code, we don't want pretty code but rather explicit code, as much as possible.

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -1,1063 +1,992 @@
 # pylint: disable=too-many-lines
+# pylint: disable=protected-access
```
bdrung, line 2: suggestion: I would move this override to the respective function since we do not want to introduce more protected access. The test should test the public part not the internal details directly.

In case you disagree, this line could be merged with the previous one.

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -1,1063 +1,992 @@
...
+    impl._sandbox_apt_cache = None
+    yield fs
+    impl.configuration = orig_conf
+    os.environ.clear()
+    os.environ.update(orig_environ)
+    shutil.rmtree(fs.workdir)
+
+
+@pytest.mark.skipif(not has_internet(), reason="online test")
+def test_install_packages_versioned(fs):
```
bdrung, line 50: pylint would complain about `fs` if invalid-name was not disabled at the top (with a todo item).

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -1,1063 +1,992 @@
...
+        os.path.join(workdir, "config"),
+    )
+    # reset internal caches between tests
+    impl._apt_cache = None
+    impl._sandbox_apt_cache = None
+    yield fs
+    impl.configuration = orig_conf
+    os.environ.clear()
+    os.environ.update(orig_environ)
+    shutil.rmtree(fs.workdir)
```
bdrung, line 46: suggestion: You could save one line by yielding `Fs` directly and `rmtree` `workdir` here.

tests/system/test_packaging_apt_dpkg.py
schopin-pro: Addressed by splitting the fixture into several smaller ones

tests/system/test_packaging_apt_dpkg.py
schopin-pro: The entire function has been scrapped :)

tests/system/test_packaging_apt_dpkg.py
schopin-pro: Replaced by individual overrides at the call sites.

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -8,1056 +8,993 @@
...
-        self.assertIn("buggerbogger", result)
-        self.assertIn("not exist", result)
 
-        # can interleave with other operations
-        dpkg = subprocess.run(
-            ["dpkg-query", "-Wf${Version}", "dash"], check=True, stdout=subprocess.PIPE
-        )
-        dash_version = dpkg.stdout.decode()
+@pytest.fixture(autouse=True)
+def environment(workdir):
```
bdrung, line 51: praise: I like that you split `environment` and `reset_impl`. The fixture code is much better to read now.

bdrung (approved):

PR #195, opened 2023-07-06
Bump the lowest Debian version to Bookworm in CI

schopin-pro:
The latest Debian stable is Bookworm, not Bullseye anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,15 @@ Please see https://wiki.ubuntu.com/Apport for more details and further links.
 The files in [doc/](./doc/) document particular details such as package hooks,
 crash database configuration, or the internal data format.
 
+Platform requirements
+---------------------
+
+The absolute minimum requirements for your distribution are Python 3.10 or
+above. Depending on which features you want enabled, various Python packages
+also need to be available.
+
+TODO: document the Python dependencies
+
 Temporarily enabling apport
 ===========================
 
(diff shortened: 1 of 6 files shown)


schopin-pro:
I greatly reduced the scope of this PR as it turns out we can't have runners that run interim Ubuntu releases, only LTS, which means our installed tests can't be run on Lunar. That means more work on #193 to fallback on old style when appropriate.

README.md
```diff
@@ -24,6 +24,15 @@ Please see https://wiki.ubuntu.com/Apport for more details and further links.
 The files in [doc/](./doc/) document particular details such as package hooks,
 crash database configuration, or the internal data format.
 
+Platform requirements
+----------------
```
bdrung, line 28:
```suggestion
---------------------
```

README.md
```diff
@@ -24,6 +24,15 @@ Please see https://wiki.ubuntu.com/Apport for more details and further links.
...
+
+The absolute minimum requirement for your distribution are Python 3.10 or
```
bdrung, line 30: question: Is that correct English? "The requirement are"

apport/packaging.py
```diff
@@ -12,50 +12,7 @@
...
-            except OSError as error:
-                errno = error.errno
-        else:
-            raise OSError(errno, f"Unable to read files {', '.join(os_release_files)}")
-
-        return ret
-
-    def freedesktop_os_release() -> typing.Dict[str, str]:
-        return _parse_os_release("/etc/os-release", "/usr/lib/os-release")
+from platform import freedesktop_os_release
```
bdrung, line 15: suggestion: For consistency we could `import platform` now and use `platform.freedesktop_os_release` later on.

PR #196, opened 2023-07-07
Increase maximum line length to 88

bdrung:
Use the default line length from black. It reduces the pressure to keep the lines short, but still keeps the lines readable. See https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length for more reasoning.


schopin-pro (approved):
I'm liking this a *lot*.

PR #197, opened 2023-07-07
refactor(apport): Make public variable pidstat private

bdrung:
Make public variable `pidstat` private and pass it around where needed.


schopin-pro (approved):

bdrung:
The codecov complains are false positives.

PR #198, opened 2023-07-08
Add support for using a gdb-sandbox to crash-digger

murraybd:
I'm working on moving the service which retraces apport crashes in Launchpad to a different server and noticed that while the existing service uses crash-digger it does not take advantage of the gdb sandbox option which I added to apport some time ago. I've added an option to crash-digger, --gdb-sandbox, which is then is then passed through as argument to apport-retrace.


murraybd:
I've tested the --gdb-sandbox option on the server too.

bdrung:
Can you add a test case for `--gdb-sandbox` to `test/integration/test_crash_digger.py` please?

PR #201, opened 2023-07-11
test: Run mypy with --disallow-incomplete-defs

bdrung:
Some functions/methods were only partially anotated with types. Complete those annotations. Use `typing.Any` for too complicated cases. This could be refined in the future.

Fix two issues found:
* `Report.add_hooks_info` could return `None` despite the docstring that this methods return a boolean. So return `False` in the `None` case.
* apport: Assert that coredump_fd is set before using it

When annotating function they should be annotated completely (for consistency). So use `--disallow-incomplete-defs` from the set of strict rules.

### DIFF
--- a/tests/integration/test_signal_crashes.py
+++ b/tests/integration/test_signal_crashes.py
@@ -808,7 +808,11 @@ class T(unittest.TestCase):
     ) -> None:
         orig_os_open = os.open
 
-        def _mocked_os_open(path, flags: int, dir_fd: Optional[int] = None) -> int:
+        def _mocked_os_open(
+            path: typing.Union[str, os.PathLike[str]],
+            flags: int,
+            dir_fd: Optional[int] = None,
+        ) -> int:
             if path == "root/run/apport.socket":
                 return orig_os_open(socket_path, flags)
             return orig_os_open(path, flags, dir_fd=dir_fd)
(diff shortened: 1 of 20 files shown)


tests/integration/test_signal_crashes.py
```diff
@@ -931,20 +933,20 @@ def create_test_process(self, command=None, uid=None, args=None):
...
-        sig=signal.SIGSEGV,
-        command=None,
-        expected_command=None,
-        uid=None,
-        expect_corefile_owner=None,
-        args=None,
+        expect_corefile: bool = False,
+        sig: int = signal.SIGSEGV,
+        command: typing.Optional[str] = None,
+        expected_command: typing.Optional[str] = None,
```
schopin-pro, line 941: suggestion (not-blocking, readability): import Optional directly rather than through typing?

This is valid for the entire PR, not just this single instance. There's a *lot* of optionals, and it's not great to read.

Might want to do this for Any, too.

schopin-pro (approved):
LGTM, only one suggestion to improve readability.

tests/integration/test_signal_crashes.py
bdrung: Okay. I will change that and see if that is better readable.

bdrung: Okay, I have changed all `typing.Optional` and `typing.Any` instanced.

PR #203, opened 2023-07-12
Deb822 prologue: refactorings

schopin-pro:
This PR contains various refactoring that will make it much easier to implement support for the APT deb822 source format.

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -38,10 +38,33 @@ import urllib.parse
 import urllib.request
 
 import apt
-import aptsources.sourceslist
+from aptsources.sourceslist import SourceEntry
 
 import apport.logging
 from apport.packaging import PackageInfo
+
+
+def _load_all_sources(apt_dir: str) -> list[SourceEntry]:
+    """Given an APT configuration directory (e.g. /etc/apt/), loads up all
+    the source data in one easy-to-consume list.
+    """
+    sources_d = os.path.join(apt_dir, "sources.list.d")
+    if os.path.exists(sources_d):
+        to_inspect = sorted([os.path.join(sources_d, f) for f in os.listdir(sources_d)])
+    else:
+        to_inspect = []
+
+    sources_old_file = os.path.join(apt_dir, "sources.list")
+    if os.path.exists(sources_old_file):
+        to_inspect[0:0] = [sources_old_file]
+
+    sources = []
+    for path in to_inspect:
+        if path.endswith(".list"):
+            with open(path, encoding="utf-8") as f:
+                sources.extend([SourceEntry(line, path) for line in f])
+
+    return sources
 
 
 class __AptDpkgPackageInfo(PackageInfo):
@@ -793,7 +816,11 @@ class __AptDpkgPackageInfo(PackageInfo):
 
             # set mirror for get_file_package()
             try:
-                self.set_mirror(self._get_primary_mirror_from_apt_sources(apt_sources))
+                self.set_mirror(
+                    self._get_primary_mirror_from_apt_sources(
+                        os.path.dirname(apt_sources)
+                    )
+                )
             except SystemError as error:
                 apport.logging.warning("cannot determine mirror: %s", str(error))
 
@@ -1334,20 +1361,18 @@ class __AptDpkgPackageInfo(PackageInfo):
         return mismatches
 
     @staticmethod
-    def _get_primary_mirror_from_apt_sources(apt_sources: str) -> str:
+    def _get_primary_mirror_from_apt_sources(apt_dir: str) -> str:
         """Heuristically determine primary mirror from an apt sources.list."""
-        with open(apt_sources, encoding="utf-8") as f:
-            for line in f:
-                source = aptsources.sourceslist.SourceEntry(line, apt_sources)
-                if source.disabled or source.invalid:
-                    continue
-                if source.type != "deb":
-                    continue
-                return source.uri
+        sources = _load_all_sources(apt_dir)
+        for source in sources:
+            if source.disabled or source.invalid:
+                continue
+            if source.type == "deb":
+                return source.uri or ""
 
         raise SystemError(
-            f"cannot determine default mirror:"
-            f" {apt_sources} does not contain a valid deb line"
+            "cannot determine default mirror:"
+            f" couldn't find configured source contains the `deb` type in {apt_dir}"
         )
 
     def _get_mirror(self):
@@ -1357,9 +1382,7 @@ class __AptDpkgPackageInfo(PackageInfo):
         configuration.
         """
         if not self._mirror:
-            self._mirror = self._get_primary_mirror_from_apt_sources(
-                "/etc/apt/sources.list"
-            )
+            self._mirror = self._get_primary_mirror_from_apt_sources("/etc/apt")
         return self._mirror
 
     def _ppa_archive_url(self, user: str, distro: str, ppa_name: str) -> str:
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -1622,6 +1622,24 @@ class __AptDpkgPackageInfo(PackageInfo):
             add_debug = ""
         return f"{ppa_line + add_debug}\ndeb-src{ppa_line[3:]}\n"
 
+    @staticmethod
+    def _find_source_file_from_origin(origin: str, src_list_d: str) -> Optional[str]:
+        """Find a possible source file matching the Origin field of a
+        given package (i.e. a PPA)"""
+        if os.path.isdir(src_list_d):
+            # check to see if there is a sources.list file for the
+            # origin, if there isn't try using a sources.list file
+            # w/o LP-PPA-
+            candidates = [os.path.join(src_list_d, f"{origin}.list")]
+            if "LP-PPA" in origin:
+                stripped = origin.strip("LP-PPA-")
+                candidates.append(os.path.join(f"{stripped}.list"))
+
+            for path in candidates:
+                if os.path.exists(path):
+                    return path
+        return None
+
     def _build_apt_sandbox(
         self, apt_root, apt_dir, distro_name, release_codename, origins
     ):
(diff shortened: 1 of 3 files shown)


tests/system/test_packaging_apt_dpkg.py
```diff
@@ -276,7 +283,8 @@ def test_install_packages_dependencies(configdir, rootdir):
 
 
 @pytest.mark.skipif(not has_internet(), reason="online test")
-def test_install_packages_system(cachedir, workdir, rootdir):
+def test_install_packages_system(cachedir, workdir, rootdir, apt_style):
+    # pylint: disable=unused-argument
```
bdrung, line 287: question: Why do you add `apt_style` as parameter here?

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -807,28 +815,31 @@ def test_use_sources_for_a_ppa(configdir, rootdir):
...
+        assert (
+            "deb http://ppa.launchpad.net/fooser/bar-ppa/ubuntu"
+            " jammy main main/debug" in sources
+        )
+        assert (
+            "deb-src http://ppa.launchpad.net/fooser/bar-ppa/ubuntu"
+            " jammy main" in sources
+        )
+    else:
+        pass
```
bdrung, line 817: How about raising NotImplemented in the else case (with a `# pragma: no cover` for the else case)?

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -868,7 +879,13 @@ def _ubuntu_archive_uri(arch=None):
     return "http://ports.ubuntu.com/ubuntu-ports"
 
 
-def _setup_foonux_config(configdir, updates=False, release="jammy", ppa=False):
+def _setup_foonux_config(
+    configdir,
+    apt_style: typing.Literal["one-line"],
```
bdrung, line 868: suggestion: How about adding the type hints for the other parameters?

tests/system/test_packaging_apt_dpkg.py
schopin-pro: Because parameters aren't optional, contrary to fixtures. I preferred having an unused argument here rather than have to parametrize every other test individually :)

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -121,16 +124,18 @@ def sandbox_ver(pkg):
...
-        "codename",
-        "sources.list",
-        "trusted.gpg.d",
-    ]
-    assert sorted(os.listdir(os.path.join(configdir, release, "armhf"))) == [
-        "sources.list",
-        "trusted.gpg.d",
-    ]
+
+    expected_content = ["sources.list.d", "trusted.gpg.d"]
```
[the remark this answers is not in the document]
schopin-pro: It's just the way the code evolved. I guess that would be the natural next step ;)

tests/system/test_packaging_apt_dpkg.py
schopin-pro: NACK. This will be changed to an actual implementation a few commits down the line, and it doesn't change anything to pass vs raise.

tests/system/test_packaging_apt_dpkg.py
schopin-pro: If you want extra type hints in my commits, you better come with a ready-to-click suggestion, my friend ;)

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -121,16 +124,18 @@ def sandbox_ver(pkg):
...
+    # old-style apt sources have the old file present
+    if apt_style == "one-line":
+        expected_content.append("sources.list")
+
+    assert set(os.listdir(os.path.join(configdir, release))) == set(
```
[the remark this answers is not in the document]
schopin-pro: Done.

tests/system/test_packaging_apt_dpkg.py
schopin-pro: Done, since I had to edit that commit anyway ;)

tests/system/test_packaging_apt_dpkg.py
bdrung: Thanks for the explanation. How about using `disable-next` here?

tests/system/test_packaging_apt_dpkg.py
bdrung: I came with https://github.com/canonical/apport/pull/201 ;)

apport/packaging_impl/apt_dpkg.py
```diff
@@ -1531,7 +1550,48 @@ def _search_contents(self, file, map_cachedir, release, arch):
                 pass
         return None
 
-    def create_ppa_source_from_origin(self, origin, distro, release_codename):
+    def _analyze_ppa_origin_string(
+        self, origin: str, distro: str
+    ) -> Optional[tuple[str, str]]:
+        assert origin.startswith("LP-PPA-")
```
bdrung, line 1556: issue: Was that change intentional? Before this commit create_ppa_source_from_origin would return `None` if origin does not start with `LP-PPA-`. With this change is will raise an assertion error. I would have expected:
```python3
if not origin.startswith("LP-PPA-"):
    return None
```

apport/packaging_impl/apt_dpkg.py
```diff
@@ -1531,7 +1550,48 @@ def _search_contents(self, file, map_cachedir, release, arch):
...
+                if index == len(components):
+                    if try_ppa:
+                        components.append("ppa")
+                        try_ppa = False
+                        index = 2
+        return None
+
+    def create_ppa_source_from_origin(
+        self, origin: str, distro: str, codename: str
+    ) -> Optional[list[SourceEntry]]:
```
bdrung, line 1594: todo: Changing the signature requires to change the docstring as well.

apport/packaging_impl/apt_dpkg.py
```diff
@@ -1670,41 +1723,19 @@ def _build_apt_sandbox(
                         "a",
                         encoding="utf-8",
                     ) as dest:
-                        dest.write(source_list_content)
-                    for line in source_list_content.splitlines():
-                        if line.startswith("#"):
-                            continue
-                        if "ppa.launchpad.net" not in line:
+                        dest.write("\n".join([str(elt) for elt in source_list_content]))
+                    for elt in source_list_content:
```
bdrung, line 1727: question: What does `elt` stand for?

bdrung (changes requested):
I finished the review of all commits. Except for some minor comment, the change looks good. I would love to see moving to pathlib (to make the code shorter and easier to read), but this is something for the future.

apport/packaging_impl/apt_dpkg.py
schopin-pro: Good point. I used to have a guard in create_ppa-source_from_origin for the prefix, but it was dropped at some point. Will be fixed in the next push.

apport/packaging_impl/apt_dpkg.py
schopin-pro: FIxed in upcoming push.

apport/packaging_impl/apt_dpkg.py
```diff
@@ -1334,20 +1357,18 @@ def _check_files_md5(sumfile):
...
-                    continue
-                return source.uri
+        sources = _load_all_sources(apt_dir)
+        for source in sources:
+            if source.disabled or source.invalid:
+                continue
+            if source.type == "deb":
+                return source.uri or ""
 
         raise SystemError(
```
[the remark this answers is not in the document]
schopin-pro: ACK. Fixed in next push.

apport/packaging_impl/apt_dpkg.py
```diff
@@ -1334,20 +1357,18 @@ def _check_files_md5(sumfile):
...
-            f"cannot determine default mirror:"
-            f" {apt_sources} does not contain a valid deb line"
+            "cannot determine default mirror:"
+            "no configured source contains the `deb` type."
```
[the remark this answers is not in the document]
schopin-pro: Fixed in next push.

apport/packaging_impl/apt_dpkg.py
```diff
@@ -803,9 +829,6 @@ def install_packages(
             ) as f:
                 self._current_release_codename = f.read().strip()
 
-        if not os.path.exists(apt_sources):
-            raise SystemError(f"{apt_sources} does not exist")
```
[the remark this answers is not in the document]
schopin-pro: Honestly, I dropped this because it's a fairly low-evidence check, and with `apt_dir` it's even worse (directory existence doesn't mean there's actual config files), and the exception itself is way too generic to be useful. If the directory doesn't exist we're going to have exceptions down the line anyway.

The fact that I didn't have to modify any test when I dropped it comforted me in that decision.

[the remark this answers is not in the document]
bdrung: Thanks for the explanation. It would be great to have that explanation in the commit message for future explanation.

apport/packaging_impl/apt_dpkg.py
```diff
@@ -1334,20 +1357,18 @@ def _check_files_md5(sumfile):
...
+        for source in sources:
+            if source.disabled or source.invalid:
+                continue
+            if source.type == "deb":
+                return source.uri or ""
 
         raise SystemError(
-            f"cannot determine default mirror:"
-            f" {apt_sources} does not contain a valid deb line"
+            "cannot determine default mirror:"
```
bdrung, line 1370:
```suggestion
            f"cannot determine default mirror:"
```
Probably a linter will complain about that.

apport/packaging_impl/apt_dpkg.py
bdrung: You also renamed `release_codename` to `codename`.

apport/packaging_impl/apt_dpkg.py
bdrung: Hm, pylint does not complain about that and that string plus f-string concatenation seems to work. So you can ignore that.

apport/packaging_impl/apt_dpkg.py
schopin-pro: Fixed (by reverting the rename)

PR #205, opened 2023-07-19
crash-digger: Add support for using a gdb-sandbox (plus test case)

bdrung:
Brian was working on moving the service which retraces apport crashes in Launchpad to a different server and noticed that while the existing service uses crash-digger it does not take advantage of the GDB sandbox option which he added to apport some time ago. He has added an option to crash-digger, --gdb-sandbox, which is then is then passed through as argument to apport-retrace.

Also add a test case for crash-digger --gdb-sandbox.


murraybd:
Thanks for picking up writing the test! I've been quite busy with some other things.

schopin-pro (approved):

PR #206, opened 2023-07-20
ui: Fix translating the web browser opening error message

bdrung:
The formatting string needs to be passed gettext before replacing the dynamic parts. This has been broken for several years.

Fixes: 300552f44369ff06647f30ab79072af6ebfc5ac9 ("merge recent changes from trunk")


schopin-pro (approved):

bdrung:
It's not trivial to add test case for this changed line (there is no existing test case for open_url). So I suggest to ignore codecov/patch in this case.

PR #211, opened 2023-07-21
apt_dpkg: Drop rootdir="/" from apt.Cache

bdrung:
When instantiating `apt.Cache` with `rootdir` set to something else than `None`, it tries to create the required directories. When running not as root and `/etc/apt/sources.list` not being present, it will fail with `PermissionError` (see https://launchpad.net/bugs/1974015).

So leave `rootdir` unset to let the cache use the system sources.list and system lists/files without trying to create `/etc/apt/sources.list`.

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -45,6 +45,7 @@ import urllib.request
 from typing import Optional
 
 import apt
+import apt_pkg
 import aptsources.sourceslist as apt_sl
 from aptsources.sourceslist import SourceEntry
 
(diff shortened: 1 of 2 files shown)


schopin-pro (approved):
LGTM, but could you please rebase on top of `main`? The compat layer has been removed.

bdrung:
The system tests are failing. That's why is split https://github.com/canonical/apport/pull/212 out of it to pin point the culprit.

julian-klode (approved):
lgtm

schopin-pro (approved):

PR #216, opened 2023-08-07
fix: Critical fix apport module error when imported "TypeError: 'type' object is not subscriptable"

jeanchristopheruel:
Critical fix
A test should be made to resolve further similar issues

```
>>> import apport
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/apport/__init__.py", line 5, in <module>
    from apport.report import Report
  File "/usr/lib/python3/dist-packages/apport/report.py", line 32, in <module>
    import apport.fileutils
  File "/usr/lib/python3/dist-packages/apport/fileutils.py", line 495, in <module>
    def get_process_environ(proc_pid_fd: int) -> dict[str, str]:
TypeError: 'type' object is not subscriptable
Error in sys.excepthook:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/apport_python_hook.py", line 72, in apport_excepthook
    from apport.fileutils import likely_packaged, get_recent_crashes
  File "/usr/lib/python3/dist-packages/apport/__init__.py", line 5, in <module>
    from apport.report import Report
  File "/usr/lib/python3/dist-packages/apport/report.py", line 32, in <module>
    import apport.fileutils
  File "/usr/lib/python3/dist-packages/apport/fileutils.py", line 495, in <module>
    def get_process_environ(proc_pid_fd: int) -> dict[str, str]:
TypeError: 'type' object is not subscriptable

Original exception was:
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/apport/__init__.py", line 5, in <module>
    from apport.report import Report
  File "/usr/lib/python3/dist-packages/apport/report.py", line 32, in <module>
    import apport.fileutils
  File "/usr/lib/python3/dist-packages/apport/fileutils.py", line 495, in <module>
    def get_process_environ(proc_pid_fd: int) -> dict[str, str]:
TypeError: 'type' object is not subscriptable

```

PR #218, opened 2023-08-07
fix: deprecated typing.Dict

jeanchristopheruel:
As discussed in #216 , [typing.Dict](https://docs.python.org/3/library/typing.html#typing.Dict) is deprecated.

### DIFF
--- a/apport/fileutils.py
+++ b/apport/fileutils.py
@@ -260,7 +260,7 @@ _LOGIN_DEFS_RE = re.compile(
 )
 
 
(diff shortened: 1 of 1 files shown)


jeanchristopheruel:
It might be a good idea to add python version requirement for installation.

bdrung:
Thanks. That can be merged once the fix for the linter (https://github.com/canonical/apport/pull/217) is merged.

bdrung (approved):

bdrung:
The linter fix was merged, but then the CI broke again and we need to merge the CI fix https://github.com/canonical/apport/pull/221 first.

PR #222, opened 2023-09-19
bash-completion: Do not use ls in completion

bdrung:
Users might set aliases for ls breaking the bash completion:

```
alias ls='ls -N --color=yes --time-style=long-iso'
apport-bug <tab>
```

Shellcheck also complains about it:

```
shellcheck -S error -s bash data/bash-completion/{apport-bug,apport-collect,apport-unpack}

In data/bash-completion/apport-bug line 8:
        for FILE in $(ls /usr/share/apport/symptoms); do
                    ^-- SC2045 (error): Iterating over ls output is fragile. Use globs.

For more information:
  https://www.shellcheck.net/wiki/SC2045 -- Iterating over ls output is fragi...
```

Ubuntu-Bug: https://launchpad.net/bugs/1850804

### DIFF
--- a/data/bash-completion/apport-bug
+++ b/data/bash-completion/apport-bug
@@ -3,17 +3,16 @@
 # get available symptoms
 _apport_symptoms ()
 {
-    local syms
-    if [ -r /usr/share/apport/symptoms ]; then 
-        for FILE in $(ls /usr/share/apport/symptoms); do
-            # hide utility files and symptoms that don't have a run() function
-            if [[ ! "$FILE" =~ ^_.* && -n $(egrep "^def run\s*\(.*\):" /usr/share/apport/symptoms/$FILE) ]]; then
-                syms="$syms ${FILE%.py}"
-            fi
-        done
-    fi
-    echo $syms
-
+    local filename path symptoms
+    for path in /usr/share/apport/symptoms/*; do
+        [[ -e "$path" ]] || continue
+        filename="${path##*/}"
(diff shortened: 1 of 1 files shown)


data/bash-completion/apport-bug
```diff
@@ -3,17 +3,16 @@
...
-            # hide utility files and symptoms that don't have a run() function
-            if [[ ! "$FILE" =~ ^_.* && -n $(egrep "^def run\s*\(.*\):" /usr/share/apport/symptoms/$FILE) ]]; then
-                syms="$syms ${FILE%.py}"
-            fi
-        done
-    fi
-    echo $syms
-
+    local filename path symptoms
+    for path in /usr/share/apport/symptoms/*; do
```
schopin-pro, line 7: question: you're removing a check here, are you sure it's safe?

schopin-pro (commented):
Overall LGTM but I do have a lingering doubt on one aspect.

data/bash-completion/apport-bug
bdrung: Which check are you referring to? Checking that `/usr/share/apport/symptoms` is present and readable? The presence check is covered by the globbing plus the following `-e "$path"` check.

schopin-pro: Thanks. I didn't remember how globbing behaved in case of non-existing path, and apparently it simply keeps the wildcard in rather than error out. All good for me.

PR #223, opened 2023-09-27
setup: replace deprecated distutils by setuptools

bdrung:
Replace deprecated distutils by setuptools:

* replace distutils.log by logging library
* drop DistUtilsExtra version check
* port build_java to setuptools

See also https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -137,7 +137,7 @@ jobs:
           bash binutils default-jdk-headless dpkg-dev gcc gdb iputils-ping kmod
           libc6-dev pkg-config python3 python3-apt python3-distutils-extra
           python3-launchpadlib python3-psutil python3-pytest python3-pytest-cov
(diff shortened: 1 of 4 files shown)


schopin-pro:
The coverage on the new java plugin is pretty poor, I wasn't expecting that!

bdrung:
The system-tests-installed should cover the new code. When running the install step in a jammy chroot, build_java and install_java are called, but they are not called in the CI. I fail to find the difference.

bdrung:
I figured it out: `entry-points` in `setup.py`/`pyproject.toml` are only going to be available for setuptools to use/discover after you install your project. So we have to wire up the java commands directly.

Do not ask me why the initial implementation did work sometimes.

schopin-pro:
OK, I'll trust you on that one, since I just know enough about setuptools to not wanting to know more about it.

PR #224, opened 2023-10-09
ui: don't rely on Dependencies to know if collect_info() has been run

schopin-pro:
Ever since we've started collecting deps info at crash file creation, that code path was essentially dead, and that resulted in the _MarkForUpload field never touched in some circumstances.

This quick workaround is to actually check for that internal field instead, as it's *only* written in collect_info(), but I think we'll need a better solution in the long term.

Fixes: ea64857890ac6cb8ad056af255cf7b706850bea7 ("feat: Already add package info during crash")
Bug-Ubuntu: https://bugs.launchpad.net/ubuntu/+source/apport/+bug/2038650

### DIFF
--- a/apport/ui.py
+++ b/apport/ui.py
@@ -484,7 +484,7 @@ class UserInterface:
             allowed_to_report = apport.fileutils.allowed_to_report()
             response = self.ui_present_report_details(allowed_to_report)
             if response.report or response.examine:
-                if "Dependencies" not in self.report:
+                if "_MarkForUpload" not in self.report:
                     self.collect_info()
 
             if self.report is None:
--- a/tests/integration/test_ui.py
+++ b/tests/integration/test_ui.py
@@ -907,6 +907,7 @@ class T(unittest.TestCase):
                 r.add_proc_info(pid)
                 r.add_user_info()
                 r.add_os_info()
+                r.add_package_info()
 
                 # generate a core dump
                 os.kill(pid, signal.SIGSEGV)


bdrung (commented):
question: `collect_info` has a "skip if we already ran" check and returns for `ProblemType` != `Crash` and `Dependencies` present. Have you tested this code path?

thought: `collect_info` sets `self.report["_MarkForUpload"] = "True"` as first step. So checking for this field means "started collection" and not "finished collection". Moving setting `_MarkForUpload` might change that.

schopin-pro:
You are indeed right that there's a change of behaviour for non-crash reports: now they might actually get uploaded! Except for Hang reports that are marked for upload on their own, any non-crash report that would come in with already-populated Dependencies would have been totally ignored. With this fix at least we get *some* data.

I haven't tested though, as I don't have the first idea how to trigger non-crash reports, and crafting bogus reports would hardly prove anything.

As far as moving `_MarkForUpload`, I say don't bother. The entire thing is an organically grown wonder of complexity, just moving a single thing around is bound to have hideous consequences. Let's just take this small win first, and then work on clearing it all up in a subsequent step.

bdrung:
Besides the `Crash` problem type there are `KernelOops`, `KernelCrash`, `Package` (for failed install/upgrade of a software package), and `Bug` (for general bug reports). We should test those as well (and ensure we have test coverage for them as well).

I'll merge this PR since it is a first step in the right direction.

bdrung (approved):

murraybd:
> Besides the `Crash` problem type there are `KernelOops`, `KernelCrash`, `Package` (for failed install/upgrade of a software package), and `Bug` (for general bug reports). We should test those as well (and ensure we have test coverage for them as well).

There is also the `RecoverableProblem` problem type which should also be considered and tested.

PR #227, opened 2023-10-23
hookutils: skip checking stat for skipped DMI files and  add unit test for apport.hookutils.attach_dmi

bdrung:
The DMI files `subsystem` and `uevent` will be always skipped. So checking their status with `os.stat` is not needed.

Add a unit test for `apport.hookutils.attach_dmi` that uses real data. This can be used as basis for fixing issues in `attach_hardware` (see https://github.com/canonical/apport/pull/226).


Chris-Peterson444 (approved):
Looks good to me!

schopin-pro (approved):

PR #233, opened 2023-10-25
launchpad: catch import error if httplib2 is missing and move tests into tests/integration/

bdrung:
The launchpadlib Python modules uses and depends on the httplib2 module. If launchpadlib is not installed, httplib2 might not be installed.

Move importing httplib2 to the guarded launchpadlib import which sets `Launchpad` to `None` in an import failure case. Then the user will get an error message presented on usage.

The Launchpad crash DB tests should not be inside the `apport` module to avoid installing them on all systems. Therefore move it to the integration tests.

Previously the Launchpad tests could be run by calling:

```
python3 apport/crashdb_impl/launchpad.py
```

Now they can be run by calling:

```
python3 -m unittest tests/integration/test_crashdb_launchpad.py
```

The tests log in to https://qastaging.launchpad.net and the user need to have bug control permissions on the Ubuntu coreutils package. They need around seven minutes to execute. Do not run them by default because a broken QA staging Launchpad instance will break them and there is no special Launchpad user for CI runs.

To enable these tests in the CI, the tests should be modified to use python3-vcr or `launchpadlib.testing.launchpad.FakeLaunchpad`. This is a task for the future.


bdrung:
The complaining codecov/patch should be ignored.

tests/integration/test_crashdb_launchpad.py
```diff
@@ -55,6 +55,10 @@ def try_to_get_from_cache(*args, **kwargs):
 
 
 @unittest.skipIf(IMPORT_ERROR, f"Python module not available: {IMPORT_ERROR}")
+@unittest.skipUnless(
+    "TEST_LAUNCHPAD" in os.environ,
+    "Need Launchpad access with bug control permission. Set TEST_LAUNCHPAD to run.",
+)
```
schopin-pro, line 61: nitpick: reverting the commits would make a better bisecting history

This is a nitpick because whoever triggers this issue probably could have narrowed their bisecting script, but having the skip guard introduced *before* we move the tests would avoid breaking integration tests in that small window.

schopin-pro (approved):
Nice find, I was surprised to find tests in that impl file :)

For future work, I think we want both cassette tests and tests that tap into the staging LP site, with the latter marked as not-blocking. That way we can actually detect whenever the LP folks break our stuff and gently nudge them before it's too late ;)

schopin-pro:
As usual, feel free to merge if you want to disregard my nitpick.

tests/integration/test_crashdb_launchpad.py
bdrung: Valid point. Changed the order of commits.

PR #235, opened 2023-10-26
test/launchpad: Some refactorings

bdrung:
* Import Report from `apport.report`
* Introduce `_create_bug_from_report()`
* set default description in `_file_bug`
* Also return web_link in `_file_bug`
* pass signal as int to `_generate_sigsegv_report`
* Move more parts into `_generate_sigsegv_report()`

Successfully tested Launchpad tests locally:
```
TEST_LAUNCHPAD=1 python3 -m unittest tests/integration/test_crashdb_launchpad.py
```

So the failing codecov/patch can be ignored.

### DIFF
--- a/tests/integration/test_crashdb_launchpad.py
+++ b/tests/integration/test_crashdb_launchpad.py
@@ -127,15 +127,6 @@ class T(unittest.TestCase):
         Return the ID.
         """
         r = self._generate_sigsegv_report()
-        r.add_package_info(self.test_package)
-        r.add_os_info()
-        r.add_gdb_info()
-        r.add_user_info()
-        self.assertEqual(r.standard_title(), "crash crashed with SIGSEGV in f()")
-
-        # add some binary gibberish which isn't UTF-8
-        r["ShortGibberish"] = ' "]\xb6"\n'
-        r["LongGibberish"] = "a\nb\nc\ne\nf\n\xff\xff\xff\n\f"
 
         return self._create_bug_from_report("SEGV", r)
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_crashdb_launchpad.py
```diff
@@ -981,8 +981,7 @@ def test_marking_python_task_mangle(self):
         # should not confuse get_fixed_version()
         self.assertEqual(self.crashdb.get_fixed_version(self.get_python_report()), None)
 
-    @staticmethod
```
schopin-pro, line 992: question (not-blocking): why was this decorator removed?

Is Python smart enough to understand that when there are no positional argument it must be a static method?

tests/integration/test_crashdb_launchpad.py
```diff
@@ -981,8 +981,7 @@ def test_marking_python_task_mangle(self):
...
-    def _generate_sigsegv_report(signal="11"):
+    def _generate_sigsegv_report(signal: int = 11) -> Report:
```
schopin-pro: praise: I'm oddly enthused by this small change

I guess I have a visceral aversion for stringly-typing?

schopin-pro (approved):
LGTM!

Just a single question, not blocking though.

tests/integration/test_crashdb_launchpad.py
bdrung: Good catch. That change is either wrong or belong to a different commit. Let me fix that.

bdrung: Fixed. That change belonged in the following commit.

PR #236, opened 2023-10-27
feat: run common hooks in alphabetical order

bdrung:
Run the common hooks in alphabetical order to make runs more predictable and to allow ordering hooks, e.g. for running a linux kernel hook after an ubuntu common hook (see bug #2018128[1]).

[1] https://launchpad.net/bugs/2018128
Signed-off-by: Benjamin Drung <benjamin.drung@canonical.com>


schopin-pro (changes requested):
I can haz tests? It'd be nice to prove that the property you claim to add is actually present.

bdrung:
Here you haz `test_add_hooks_info` extended to test the execution order. I tested that this extended test case will fail with the old code.

schopin-pro (approved):

PR #237, opened 2023-10-27
refactor: add _write_key_and_binary_value_compressed_and_encoded_to_file and add type hints

bdrung:
Carve out `_write_key_and_binary_value_compressed_and_encoded_to_file` from `ProblemReport.write` to get this method clean and readable. Also add type hints. mypy will complain about `crc`:
> Incompatible types in assignment (expression has type "None", variable has type "int")

Introduce the boolean variable `write_aborted` to address mypy and make the code more readable.

### DIFF
--- a/problem_report.py
+++ b/problem_report.py
@@ -483,6 +483,7 @@ class ProblemReport(collections.UserDict):
         file.write(base64.b64encode(gzip_header))
         file.write(b"\n ")
         crc = zlib.crc32(b"")
+        write_aborted = False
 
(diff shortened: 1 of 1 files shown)


schopin-pro:
Ugh, is coverage calculation drunk again? At least this time it's in the right direction, I guess...

schopin-pro (approved):

bdrung:
Only the main coverage of 73.66% seems to be off (probably the CI runs on the main branch were not finished).

PR #239, opened 2023-10-30
feat: set signal name during crash collection

bdrung:
Instead of hard-coding only five signal names, determine the signal name using the Python `signal` standard library.

Signal numbers vary by architecture. So include the signal name during crash collection to avoid getting the wrong name when reprocessing on a different architecture.


schopin-pro (approved):
I like this, small QoL improvement for us :)

PR #249, opened 2023-11-04
problem_report: rename "key and value" to item

bdrung:
The key and value pairs in Python dicts are called items. So rename the "key and value" part of methods to "item".

Maybe we could drop `_to_file` from the methods to make the function name shorter.


schopin-pro (changes requested):
I would indeed drop the `_to_file`

bdrung:
Okay, dropped the `_to_file`.

schopin-pro (approved):
Yup, dropping to_file is indeed much nicer \o/

PR #252, opened 2023-11-07
ui: always return size as int

bdrung:
Always return the complete or reduces size as `int`. Round the guessed size from float to int.


bdrung:
This merge request is obsoleted by https://github.com/canonical/apport/pull/257.

PR #253, opened 2023-11-16
test/apport-valgrind: do not limit diff size

bdrung:
Set `maxDiff` to `None` to see full diff output for better debugging bugs like https://launchpad.net/bugs/2043713


schopin-pro (approved):
Fair enough.

PR #254, opened 2023-11-29
pylint: load more pylint plugins

bdrung:
Load more pylint plugins that currently do not complain. This includes useful plugins like `no_self_use`. The `typing` plug-in detects using the deprecated `typing.Tuple`.

### DIFF
--- a/tests/system/test_packaging_apt_dpkg.py
+++ b/tests/system/test_packaging_apt_dpkg.py
@@ -933,7 +933,7 @@ def _ubuntu_archive_uri(arch=None):
     """Return archive URI for the given architecture."""
     if arch is None:
         arch = impl.get_system_architecture()
-    if arch in ("amd64", "i386"):
+    if arch in {"amd64", "i386"}:
         return "http://archive.ubuntu.com/ubuntu"
     return "http://ports.ubuntu.com/ubuntu-ports"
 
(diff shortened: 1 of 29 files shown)


tests/system/test_packaging_apt_dpkg.py
```diff
@@ -272,7 +272,7 @@ def test_install_packages_unversioned(configdir, cachedir, rootdir, apt_style):
     assert "coreutils 8.32-4.1ubuntu1" in pkglist
     assert "coreutils-dbgsym 8.32-4.1ubuntu1" in pkglist
     assert "tzdata 2022a-0ubuntu1" in pkglist
-    assert len(pkglist), 3 == str(pkglist)
```
schopin-pro, line 275: thought: I'm not sure what this assertion means, to be honest.

schopin-pro (approved):
LGTM, but the weird assertion merits looking into (in a separate PR)

tests/system/test_packaging_apt_dpkg.py
bdrung: You might want to ask the author of commit d5f1f43682d2c640dcc9af68f793ba28b5a58b40. :wink: 

The code before was:
```
self.assertEqual(len(pkglist), 3, str(pkglist))
```

schopin-pro: Mmmh, that must have fallen through the cracks when whoever was assigned as reviewer read that diff :wink: .

This would be so much better to do

```py
assert set(pkglist) == {"coreutils 8.32-4.1ubuntu1", "coreutils-dbgsym 8.32-4.1ubuntu1", "coreutils-dbgsym 8.32-4.1ubuntu1"}
```

bdrung: Yes, please create a merge request for that.

PR #262, opened 2023-12-05
problem_report: fix setting name of CompressedValue

bdrung:
The name of the compressed value is the second parameter of `CompressedValue`. The first value is the decoded value.


problem_report.py
```diff
@@ -212,7 +212,7 @@ def load(self, file, binary=True, key_filter=None):
                 value = value.strip()
                 if value == b"base64":
                     if binary == "compressed":
-                        value = CompressedValue(key.encode())
+                        value = CompressedValue(name=key)
```
schopin-pro, line 215: issue: this doesn't smell right!

Either we have a gap in testing this feature, or setting the name is actually useless, but either way a deeper investigation is warranted.

schopin-pro (changes requested):
Short, sweet, simple... and incomplete.

problem_report.py
bdrung: Added a test case for it.

schopin-pro (approved):

PR #266, opened 2023-12-08
add --minimum-java-release option to the build command

vpa1977:
In order to compile with default Java 21 we need to use release 8 instead of 7. 

To avoid changing java compilation class in the future it would be better to make it the minimum java release customisable via a command line option. 

This PR implements `--minimum-java-release=` command line argument for the build command.

With Java 11+ we should be using --release option rather than -source/-target to avoid spurious runtime-related errors. We should fix it as a separate PR after this one.

### DIFF
--- a/setuptools_apport/java.py
+++ b/setuptools_apport/java.py
@@ -16,10 +16,15 @@ class build_java(Command):
     """Compile Java components of Apport"""
 
     description = __doc__
-    user_options: list[tuple[str, str, str]] = []
+    user_options = [("minimum-java-release=", "r", "Specify minimum Java release.")]
+
+    def __init__(self, dist: Distribution, **kwargs: dict[str, typing.Any]) -> None:
+        Command.__init__(self, dist, **kwargs)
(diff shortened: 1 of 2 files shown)


setuptools_apport/java.py
```diff
@@ -16,10 +16,17 @@ class build_java(Command):
     """Compile Java components of Apport"""
 
     description = __doc__
-    user_options: list[tuple[str, str, str]] = []
+    user_options: list[tuple[str, str, str]] = [
+        ("minimum-java-release=", "r", "Specify minimum Java release.")
```
bdrung, line 20: question: What is the second parameter used for? I only found usages where the second one was set to `None`.

vpa1977: I think -r is a shorthand option (if I understand the question correctly =) )

bdrung: Yes. That's it. I tested it with `setup.py -r 9`. So I learned something new today.

bdrung (approved):

PR #278, opened 2023-12-11
ui: remove skipping crashes that happened during logout

bdrung:
Ubuntu >= 22.04 uses only cgroup2 (Ubuntu 20.04 used both and Ubuntu 18.04 only cgroup version 1). `Report.get_logind_session` will fail to determine the logind session and always return `None` if only cgroup2 is used.

`get_logind_session` searches for cgroup lines with "name=systemd:" and then extracts the session from "/session-". This worked in Ubuntu 16.04 but started to fail in Ubuntu 18.04.

Content of `/proc/self/cgroup` on Ubuntu 16.04 (xenial):

```
12:cpuset:/
11:perf_event:/
10:pids:/user.slice/user-1000.slice
9:blkio:/
8:hugetlb:/
7:net_cls,net_prio:/
6:cpu,cpuacct:/
5:devices:/user.slice
4:freezer:/
3:rdma:/
2:memory:/
1:name=systemd:/user.slice/user-1000.slice/session-c1.scope
```

Content of `/proc/self/cgroup` on Ubuntu 18.04 (bionic):

```
12:freezer:/
11:net_cls,net_prio:/
10:memory:/
9:rdma:/
8:cpu,cpuacct:/
7:devices:/user.slice
6:perf_event:/
5:cpuset:/
4:pids:/user.slice/user-1000.slice/user@1000.service
3:hugetlb:/
2:blkio:/
1:name=systemd:/user.slice/user-1000.slice/user@1000.service/gnome-terminal-server.service
0::/user.slice/user-1000.slice/user@1000.service/gnome-terminal-server.service
```

Content of `/proc/self/cgroup` on Ubuntu 22.04 (jammy):

```
0::/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-0c3f8b02-b112-4cab-accc-764cbb223bac.scope
```

So crashes that happened during logout haven't been skipped for over five years and no one filed a bug report for it. So just remove the broken code.

Bug: https://launchpad.net/bugs/2043393

### DIFF
--- a/apport/report.py
+++ b/apport/report.py
@@ -680,7 +680,7 @@ class Report(problem_report.ProblemReport):
         extraenv: (Iterable[str] | None) = None,
     ) -> None:
         # TODO: Split into smaller functions/methods
-        # pylint: disable=too-many-branches,too-many-statements
+        # pylint: disable=too-many-branches
         """Add /proc/pid information.
 
         If neither pid nor self.pid are given, it defaults to the process'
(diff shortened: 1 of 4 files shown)


schopin-pro (approved):
The logic is sound. Worst case scenario we can always revert this and *try* to fix it if people start complaining.

PR #279, opened 2023-12-13
coverage: automatically exclude skipTest lines

schopin-pro:
Since we're aiming at enabling as many tests as possible on CI, it makes sense to remove those lines from coverage since they would fall out of the covered paths as our CI gets better.


schopin-pro:
Closing this out as it's going to be a PITA as long as we support Jammy.

PR #281, opened 2024-01-05
feat: support reading report from systemd-coredump

bdrung:
There can only be one system core dump handler on a Linux system and Apport is the Ubuntu default core dump handler. Besides Apport other tools offer coredump handler, in particular systemd-coredump. Due to Apport being installed on all Ubuntu system by default none of these alternate system can be installed.

Implement converting a coredump from systemd-coredump into an Apport problem report.

Bug-Ubuntu: https://launchpad.net/bugs/2048136

### DIFF
--- a/apport/report.py
+++ b/apport/report.py
@@ -14,6 +14,7 @@
 # pylint: disable=invalid-name
 
 import atexit
+import datetime
 import errno
 import fnmatch
 import glob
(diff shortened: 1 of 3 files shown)


schopin-pro (approved):
LGTM, I like the overall architecture, this was easy to follow :)

PR #282, opened 2024-01-05
ci: also measure branch coverage

bdrung:
In addition to the usual statement coverage, also collect branch coverage measurement to ensure that those code path are also tested.


schopin-pro (commented):
I don't feel strongly either way, so let's merge this, we can always revert it later anyway!

PR #284, opened 2024-01-09
test: add test_crash_suid_dumpable_debug

bdrung:
To increase the code coverage add a test for
`/proc/sys/fs/suid_dumpable` set to 1 ("debug") and crashing a set-user-ID (suid) program.

### DIFF
--- a/tests/integration/test_signal_crashes.py
+++ b/tests/integration/test_signal_crashes.py
@@ -41,6 +41,7 @@ from tests.paths import get_data_directory, local_test_environment
 APPORT_PATH = get_data_directory() / "apport"
 apport_binary = import_module_from_file(APPORT_PATH)
 
+MAIL_UID = 8
 test_package = "coreutils"
 test_source = "coreutils"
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_signal_crashes.py
```diff
@@ -680,7 +680,18 @@ def test_crash_setuid_keep(self):
 
     @unittest.skipUnless(os.path.exists("/bin/ping"), "this test needs /bin/ping")
     @unittest.skipIf(os.geteuid() != 0, "this test needs to be run as root")
-    def test_crash_setuid_drop(self):
+    def test_crash_suid_dumpable_debug(self):
+        """Report generation for setuid program which drops root."""
+        # run ping as user "mail"
+        resource.setrlimit(resource.RLIMIT_CORE, (-1, -1))
```
schopin-pro, line 686: nitpick: the comment doesn't match the line.

tests/integration/test_signal_crashes.py
```diff
@@ -680,7 +680,18 @@ def test_crash_setuid_keep(self):
...
+
+        # if a user can crash a suid root binary, it should not create
+        # core files if /proc/sys/fs/suid_dumpable is set to 1 ("debug")
+        self.do_crash(command="/bin/ping", args=["127.0.0.1"], uid=8, suid_dumpable=1)
```
schopin-pro, line 690: question: is there a guarantee that UID 8 is `mail` and even present on the test system?

bdrung:
I checked what systemd-coredump is doing: It writes the core file as root and does not allow the user to read it in case of suid processes:
```
        /* We don't allow users to read coredumps if the uid or capabilities were changed. */
        if (!allow_user)
                return 0;
```

tests/integration/test_signal_crashes.py
bdrung: base-files relies on the `mail` user to be present. So it will be on the test systems. On Debian/Ubuntu `mail` is always UID 8, but I failed to find the component that is responsible for the initial content of `/etc/passwd`.

tests/integration/test_signal_crashes.py
```diff
@@ -41,6 +41,7 @@
 APPORT_PATH = get_data_directory() / "apport"
 apport_binary = import_module_from_file(APPORT_PATH)
 
+MAIL_UID = 8
```
schopin-pro, line 44: praise: thanks for that, makes it easier to understand

Also should make it easier to change it down the line should we come upon circumstances where the mail UID is *not* 8.

tests/integration/test_signal_crashes.py
```diff
@@ -680,7 +680,22 @@ def test_crash_setuid_keep(self):
...
+
+        # if a user can crash a suid root binary, it should not create
+        # core files if /proc/sys/fs/suid_dumpable is set to 1 ("debug")
+        self.do_crash(
+            command="/bin/ping", args=["127.0.0.1"], uid=MAIL_UID, suid_dumpable=1
+        )
+
+    @unittest.skipUnless(os.path.exists("/bin/ping"), "this test needs /bin/ping")
+    @unittest.skipIf(os.geteuid() != 0, "this test needs to be run as root")
+    def test_crash_setuid_(self):
```
schopin-pro, line 698: issue: you dropped the `drop` part of the test name?

schopin-pro (changes requested):
Almost there! It seems there was a typo or something?

PR #292, opened 2024-02-09
fix: use context manager when manipulating GzipFiles (LP: #2051512)

schopin-pro:
Those objects are technically file-like objects, with the associated semantics. In particular, starting with Python 3.12 the class using internal buffering, which means write() calls won't necessarily directly be reflected in the underlying storage object.

For the sake of consistency, the context manager approach has been generalized to read() calls as well, except when we could get away with removing the use of GzipFile altogether.


schopin-pro:
I'm not going to write new tests for this, the problem *was* caught by tests in the first place ! ;)

bdrung (approved):

PR #293, opened 2024-02-09
ci: upgrade to codecov/codecov-action@v4

bdrung:
GitHub CI warns: "Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, codecov/codecov-action@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/."

So upgrade codecov/codecov-action@v3 to codecov/codecov-action@v4. Explicitly set `SKIP_ONLINE_TESTS=1` for the skip tests, because installing ca-certificates causes the skip tests to have their Internet check succeed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,6 +8,9 @@ on:  # yamllint disable-line rule:truthy
 env:
   DEBIAN_FRONTEND: noninteractive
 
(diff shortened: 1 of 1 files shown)


schopin-pro:
Forcing it through, the failure is only because of rate-limiting.

PR #294, opened 2024-02-13
dpkg: memoize the get_system_architecture method

schopin-pro:
It's rather costly, shelling out to dpkg and all.

This had the consequence of having to rename the `cache` variable into `apt_cache`, but it might be for the better anyway as there are already other caches around.

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 
 import contextlib
 import datetime
+import functools
 import glob
 import gzip
 import hashlib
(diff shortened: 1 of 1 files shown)


bdrung (commented):
question: Why not `import functools` and then use `functools.cache`? Then you do not need the rename and it is more clear where `cache` comes from.

schopin-pro:
I thought about it, but IMO the rename is worthwhile on its own, so I went for it.

bdrung:
IMO splitting those two changes into separate commits would be good. My personal preference would be to use `functools.cache` regardless of the variable rename.

bdrung (approved):

schopin-pro:
I guess it makes sense. I split off the renaming into its own PR.

bdrung (approved):

PR #295, opened 2024-02-13
Rework apport-retrace to handle unbound crashid (LP: #2051512)

schopin-pro:
A small amount of refactoring in apport-retrace to understand the provenance of `crashid` and make it clear for both tooling and readers that it has a valid value when trying to push the retrace out to the servers.

Depends on #296 to make mypy happy.

### DIFF
--- a/bin/apport-retrace
+++ b/bin/apport-retrace
@@ -26,12 +26,14 @@ import tempfile
 import termios
 import tty
 import zlib
+from argparse import Namespace
 from gettext import gettext as _
 
 import apport
 import apport.fileutils
 import apport.sandboxutils
-from apport.crashdb import get_crashdb
+from apport import Report
+from apport.crashdb import CrashDatabase, get_crashdb
 
 
 def parse_args():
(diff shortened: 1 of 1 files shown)


schopin-pro:
To help review and to make sure we can integrate the actual fix earlier, I ended up splitting the PR in 3, with this one being the middle piece. It's still not covered by tests, but review should make it obvious that the changes are mostly cosmetics at that point.

I'd have hoped that the type hint one would have been completely autonomous, but sadly mypy needs it to understand that `apport.fatal()` is, well, fatal.

bin/apport-retrace
```diff
@@ -389,14 +384,29 @@ Thank you for your understanding, and sorry for the inconvenience!
                 sys.exit(0)
             else:
                 raise
-
-        crashid = options.report
-        options.report = None
```
schopin-pro, line 394: For the record, this was the line that allowed the use of crashid when protected by `if not options.report` further down.

bdrung: That was quickly clear to me when I just looked at the commits individually.

bin/apport-retrace
```diff
@@ -637,8 +647,8 @@ Thank you for your understanding, and sorry for the inconvenience!
                     if "Stacktrace" in report:
                         crashdb.update_traces(crashid, report)
                         apport.log(
-                            "New attachments uploaded to crash database LP: #"
-                            + crashid,
+                            f"New attachments uploaded to crash database"
+                            f"LP: #{crashid}",
```
bdrung, line 651: nitpick: That line wrapping should be done in the first commit where you change that line.

bdrung (commented):
typo: `falsey` in the second commit message.

praise: Thanks for reducing this PR to the needed changes (since there are no test cases for this binary).

schopin-pro:
All resolved, including the typo in the commit message (s/falsey/falsy)

bin/apport-retrace
bdrung, line 651:
```suggestion
                            f" LP: #{crashid}",
```

bdrung:
Shouldn't it be `falsely` instead?

schopin-pro:
Nope, I stand by "falsy", as in "evaluate to False where a boolean is expected". It's a thing, I swear ;-)

bin/apport-retrace
schopin-pro: *sigh*
Fixed.

bdrung (approved):

PR #297, opened 2024-02-14
Some apport-retrace refactoring

schopin-pro:
A couple of commits that split out some code out of apport-retrace's main() and into helper functions.

Depends on #295


schopin-pro:
Marking as Draft since there are no tests to back it up.

PR #299, opened 2024-02-14
ci: run integration tests also on Debian

bdrung:
The libc-bin package suggests `manpages` on Ubuntu, but recommends it on Debian. If the recommended packages are not installed, the test `test_get_dependencies_depends_only` will fail on Debian:

```
_____________________ T.test_get_dependencies_depends_only _____________________

self = <tests.integration.test_packaging_apt_dpkg.T testMethod=test_get_dependencies_depends_only>

    def test_get_dependencies_depends_only(self):
        """get_dependencies() on package with Depends only."""
        d = impl.get_dependencies("libc-bin")
        self.assertGreaterEqual(len(d), 1)
        for dep in d:
>           self.assertTrue(impl.get_version(dep))

tests/integration/test_packaging_apt_dpkg.py:114:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <apport.packaging_impl.apt_dpkg.__AptDpkgPackageInfo object at 0x7f659f504fd0>
package = 'manpages'

    def get_version(self, package):
        """Return the installed version of a package."""
        pkg = self._apt_pkg(package)
        inst = pkg.installed
        if not inst:
>           raise ValueError(f"package {package} does not exist")
E           ValueError: package manpages does not exist

apport/packaging_impl/apt_dpkg.py:266: ValueError
```

Use the essential package `sysvinit-utils` which only depends on `libc6` in Debian and Ubuntu. This package does not recommend/suggest anything.

With this fix, the integration tests can also be run on Debian.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,7 +39,7 @@ jobs:
       - name: Run linter tests
         run: tests/run-linters
 
-  unit:
+  unit-and-integration:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


schopin-pro (approved):
LGTM

PR #301, opened 2024-02-14
ui: Drop forking in open_url and show error message if webbrowser.open() fails

bdrung:
The `new` parameter in `webbrowser.open` takes an integer and not a boolean: "If `new` is 0, the url is opened in the same browser window if possible. If `new` is 1, a new browser window is opened if possible. If `new` is 2, a new browser page (“tab”) is opened if possible."

run_as_real_user uses `subprocess.run` and `webbrowser.open` which are non-blocking. So forking can be dropped.

`webbrowser.open` returns a boolean. If openining a browser fails (e. g. in a minimal chroot), `webbrowser.open` will return `False`. In this case `UserInterface.open_url` silently does nothing. Print/show an error message in case `webbrowser.open` returns `False`.

### DIFF
--- a/apport/ui.py
+++ b/apport/ui.py
@@ -1733,7 +1733,7 @@ class UserInterface:
                 run_as_real_user(["xdg-open", url], get_user_env=True)
             except OSError:
                 # fall back to webbrowser
-                webbrowser.open(url, new=True, autoraise=True)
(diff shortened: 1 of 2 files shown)


schopin-pro (approved):
I like this, especially the third commit (I've had issues with this exact silent failure before)

PR #307, opened 2024-04-03
test: fix date in test_crash_apport_from_systemd_coredump

bdrung:
test_crash_apport_from_systemd_coredump fails if the current day of the month contains only one digit:

```
>       self.assertEqual(dict(report), dict(expected_report))
E       AssertionError: {'Pro[16 chars]h', 'Architecture': 'amd64', 'Date': 'Wed Apr [4740 chars]'no'} != {'Pro[16 chars]h', 'Date': 'Wed Apr 03 15:32:15 2024', 'ProcC[4740 chars]'no'}
E         {'Architecture': 'amd64',
E       -  'Date': 'Wed Apr  3 15:32:15 2024',
E       ?                   ^
E
E       +  'Date': 'Wed Apr 03 15:32:15 2024',
E       ?                   ^
E
```

### DIFF
--- a/tests/integration/test_signal_crashes.py
+++ b/tests/integration/test_signal_crashes.py
@@ -821,7 +821,7 @@ class T(unittest.TestCase):
 
             now = datetime.datetime.now(tz=datetime.timezone.utc)
             expected_report = apport.report.Report(
-                date=now.strftime("%a %b %d %H:%M:%S %Y")
+                date=now.strftime("%a %b %e %H:%M:%S %Y")
             )
             expected_report.pid = pid
             expected_report.add_proc_info()


tests/integration/test_signal_crashes.py
```diff
@@ -821,7 +821,7 @@ def test_crash_apport_from_systemd_coredump(self, reader_mock: MagicMock) -> Non
 
             now = datetime.datetime.now(tz=datetime.timezone.utc)
             expected_report = apport.report.Report(
-                date=now.strftime("%a %b %d %H:%M:%S %Y")
+                date=now.strftime(f"%a %b {now.day:>2} %H:%M:%S %Y")
```
schopin-pro, line 824: suggestion: use width specifier as documented in strftime(3)
```suggestion
                date=now.strftime("%a %b %2d %H:%M:%S %Y")
```

As hinted at by the Python documentation, `time.strftime()` is a thin wrapper around the libc implementation, and glibc supports width specifiers.

schopin-pro (changes requested):
I think we can avoid mixing fstring into that mess :)

tests/integration/test_signal_crashes.py
bdrung: That doesn't work: `%d` already resolves to the date with `0` prefixed. But `%e` works (but is not mentioned in the Python documentation).

schopin-pro (approved):

PR #310, opened 2024-04-05
apt_dpkg: fix heuristic for determining primary mirror with DEB822 sources

bdrung:
On systems with DEB822 sources, determining the primary mirror will probably fail when other sources (like PPAs) are added. On my system with DEB822 sources the `test_sandbox_cache_options` test case fails with `urllib.error.HTTPError: HTTP Error 404: Not Found` because it tries to query the PPA as primary Ubuntu mirror:
`https://ppa.launchpadcontent.net/<user>/<ppa>/ubuntu/dists/mantic/Contents-amd64.gz` does not exist.

So check for the `.sources` file named after the distribution (e.g. `ubuntu.sources` for Ubuntu and `debian.sources` for Debian) first for the primary mirror.

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -44,6 +44,7 @@ import time
 import urllib.error
 import urllib.parse
 import urllib.request
+from collections.abc import Iterator
 
 import apt
 import apt_pkg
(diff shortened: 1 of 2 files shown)


apport/packaging_impl/apt_dpkg.py
```diff
@@ -81,14 +82,28 @@ def _parse_deb822_sources(source: str) -> list[apt_sl.Deb822SourceEntry]:
         return sections
 
 
+def _order_sources(apt_sources: list[str]) -> Iterator[str]:
+    """Order APT sources files heuristically to get the primary sources first."""
+    sources = set(apt_sources)
+    for expected_primary in ("debian.sources", "ubuntu.sources"):
```
schopin-pro, line 88: suggestion: check if the system is Ubuntu or Debian (lsb_release ?) rather than check for *both* sources

schopin-pro (changes requested):

apport/packaging_impl/apt_dpkg.py
bdrung: Good idea. implemented.

apport/packaging_impl/apt_dpkg.py
```diff
@@ -81,14 +82,29 @@ def _parse_deb822_sources(source: str) -> list[apt_sl.Deb822SourceEntry]:
         return sections
 
 
+def _order_sources(apt_sources: list[str]) -> Iterator[str]:
+    """Order APT sources files heuristically to get the primary sources first."""
+    sources = set(apt_sources)
+    info = platform.freedesktop_os_release()
+    expected_primary = f"{info['ID']}.sources"
```
schopin-pro, line 89: praise: I like this solution, that's rather elegant :)

schopin-pro (approved):

PR #311, opened 2024-04-05
apt_dpkg: split _search_contents into smaller methods

bdrung:
Split `__AptDpkgPackageInfo._search_contents()` into smaller methods to make the code more readable and modifiable.

I did this refactoring to find the test failure, but then I found the failure elsewhere: https://github.com/canonical/apport/pull/310

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -1487,9 +1487,141 @@ class __AptDpkgPackageInfo(PackageInfo):
             " without install_packages()"
         )
 
+    def _fetch_contents_file(
+        self, contents_filename: str, mtime: float | None, dist: str, arch: str
+    ) -> bool:
+        update = False
+        url = f"{self._get_mirror()}/dists/{dist}/Contents-{arch}.gz"
+        if mtime:
+            # HTTPConnection requires server name e.g.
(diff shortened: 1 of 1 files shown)


apport/packaging_impl/apt_dpkg.py
```diff
@@ -1485,130 +1619,14 @@ def _search_contents(self, file, map_cachedir, release, arch):
...
-                        else:
-                            continue
-                        if path in contents_mapping:
-                            if package == contents_mapping[path]:
-                                continue
-                            # if the package was updated use the update
-                            # b/c everyone should have packages from
-                            # -updates and -security installed
-                        contents_mapping[path] = package
+        contents_mapping = self._get_contents_mapping(map_cachedir, release, arch)
```
schopin-pro, line 1623:
```suggestion
        file2pkg = self._get_contents_mapping(map_cachedir, release, arch)
```

We don't really care that it comes from a Contents file once we have the mapping, and that name makes its contents (heh!) explicit.

schopin-pro (changes requested):

apport/packaging_impl/apt_dpkg.py
bdrung: Renamed and also adjusted some of the helper methods to use that name.

PR #313, opened 2024-04-12
remove dependency on python-requests

bdrung:
Importing `requests` uses about 10 MB of additional memory and costs about 100 milliseconds. My personal motivation is more to push backwards against the increasing size and complexity of the Ubuntu base system, or alternatively, to reduce the number of projects that constitute critical infrastructure for Linux.

`urllib.request` raises an exception for anything other than a 20x by default (this is `urllib.request.HTTPErrorHandler`).


slingamn:
Thanks for doing this! The code change looks good to me.

PR #335, opened 2024-06-07
test: propery mock os.uname calls

bdrung:
Use `unittest.mock.patch` to mock calls to `os.uname` instead of manually replacing and restoring `os.uname`.


schopin-pro:
Tests are mad!

schopin-pro (approved):

PR #337, opened 2024-06-08
test: convert workdir to a pathlib.Path

bdrung:
Using `pathlib.Path` is more readable.


schopin-pro (commented):
I think this half-baked state is not more readable, now we have some dir attributes that are Paths while others are strings?

bdrung:
Changing the paths in the apport module could break the API. So let's keep it as it is.

PR #338, opened 2024-06-08
test: move generate_sleep_crash to helper

bdrung:
Move `generate_sleep_crash` to `tests/helper.py` to make it usable by system test cases for `apport-retrace`.


bdrung:
Not needed any more after reworking apport-retrace system tests.

PR #341, opened 2024-06-11
report: Fix retracing crashes on Ubuntu noble

bdrung:
Retracing crashes from Ubuntu 24.04 (noble) can fail with (test case `test_retrace_system_sandbox_gdb_sandbox`):

```
ERROR: [Errno 2] No such file or directory: '/tmp/apport_sandbox_8s5bjkjo/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2'
```

Let `subprocess.run` find the correct path to `ld-linux-x86-64.so.2` by letting it search through all library paths.

Add system tests for apport-retrace: One set of tests for retracing the same release and one set of tests for retracing crashes of a different release (in this case jammy).

Bug-Ubuntu: https://launchpad.net/bugs/2067120


bdrung:
Creating a new merge request does not help.

PR #348, opened 2024-06-13
Add type hints to UserInterface class

bdrung:
Add type hints to all occurrences of the `UserInterface` class.

mypy on jammy has false positives:

```
kde/apport-kde:441: error: Argument 1 to "information" of "QMessageBox" has incompatible type "None"; expected "QWidget"
kde/apport-kde:444: error: Argument 1 to "information" of "QMessageBox" has incompatible type "None"; expected "QWidget"
kde/apport-kde:535: error: Argument 1 to "question" of "QMessageBox" has incompatible type "None"; expected "QWidget"
```

Do not install `python3-pyqt5` on jammy to avoid those false positives.


bdrung:
The CI is green except that upload-to-codecov is failing.

PR #354, opened 2024-06-13
report: use library search order from ld.so

bdrung:
Use the same library search order from `ld.so`:

```
$ ld.so --help
[...]
Shared library search path:
  (libraries located via /etc/ld.so.cache)
  /lib/x86_64-linux-gnu (system search path)
  /usr/lib/x86_64-linux-gnu (system search path)
  /lib (system search path)
  /usr/lib (system search path)
```

### DIFF
--- a/apport/report.py
+++ b/apport/report.py
@@ -286,6 +286,21 @@ def _get_gdb_path(gdb_sandbox: str | None, same_arch: bool) -> str:
     return gdb_path
 
 
+def _get_ld_search_paths(ld_so: str = "ld.so") -> list[str]:
+    """Query the given ld.so binary for the shared library search paths."""
+    diagnostics = subprocess.run(
+        [ld_so, "--list-diagnostics"], check=True, stdout=subprocess.PIPE, text=True
+    )
+    # Example matching line: path.system_dirs[0x0]="/lib/x86_64-linux-gnu/"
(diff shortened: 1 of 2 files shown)


apport/report.py
```diff
@@ -1905,9 +1905,9 @@ def gdb_command(
             ]
             if gdb_sandbox:
                 ld_lib_path = (
-                    f"{gdb_sandbox}/lib"
                     f":{gdb_sandbox}/lib/{native_multiarch}"
                     f":{gdb_sandbox}/usr/lib/{native_multiarch}"
+                    f"{gdb_sandbox}/lib"
```
schopin-pro, line 1910: issue: Rather than a hardcoded list, shouldn't we actually *query* the local ld.so to get the search order?

schopin-pro (changes requested):

apport/report.py
bdrung: You asked for it. Now you get a bigger merge request!

bdrung:
Problem with the implementation: jammy does not ship `/usr/bin/ld.so`.

schopin-pro:
I didn't even remember that `/usr/bin/ld.so` was a thing. Maybe just use `/lib/$triplet/ld.*\.so\..*`?

PR #358, opened 2024-06-21
test: add type hints to unit tests

bdrung:
Add type hints to all unit tests.

### DIFF
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -52,7 +52,7 @@ def import_module_from_file(path: pathlib.Path) -> Any:
     return module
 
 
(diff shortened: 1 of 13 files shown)


tests/helper.py
```diff
@@ -52,7 +52,7 @@ def import_module_from_file(path: pathlib.Path) -> Any:
     return module
 
 
-def pidof(program):
+def pidof(program: str) -> set[int]:
```
schopin-pro, line 55: thought: this really should be renamed `pidf_of()` at some point.

schopin-pro (approved):

PR #359, opened 2024-06-28
Fix retrace tests on non-amd64 (FR-8039)

schopin-pro:
There are a few assumptions made by the test code that only are true on amd64 :)

See https://bugs.launchpad.net/ubuntu/+source/apport/+bug/2069815

Note that we'll probably want to install gdb-multiarch in debian/tests/control in the packaging.

### DIFF
--- a/tests/system/test_apport_retrace.py
+++ b/tests/system/test_apport_retrace.py
@@ -201,6 +201,38 @@ def test_retrace_system_sandbox(
 
 
 @pytest.mark.skipif(not has_internet(), reason="online test")
+@pytest.mark.skipif(
+    impl.get_system_architecture() == "amd64",
+    reason="GDB sandbox is only available on amd64",
+)
+def test_retrace_system_sandbox_gdb_sandbox_nonamd64(
+    workdir: pathlib.Path, module_cachedir: pathlib.Path, divide_by_zero_crash: str
+) -> None:
+    """Refuse to retrace a crash in a non-amd64 system sandbox with a GDB sandbox."""
+    retraced_report_filename = workdir / "retraced.crash"
+    env = os.environ | local_test_environment()
+    cmd = [
+        "apport-retrace",
+        "-v",
+        "-o",
+        str(retraced_report_filename),
+        "--sandbox",
+        "system",
+        "--gdb-sandbox",
+        "--cache",
+        str(module_cachedir),
+        divide_by_zero_crash,
+    ]
+    ret = subprocess.run(cmd, check=False, env=env, capture_output=True, text=True)
+    assert ret.returncode == 3
+    assert "gdb sandboxes are only implemented for amd64 hosts" in ret.stderr
+
+
+@pytest.mark.skipif(not has_internet(), reason="online test")
+@pytest.mark.skipif(
+    impl.get_system_architecture() != "amd64",
+    reason="Testing the GDB sandbox erroring out on non-AMD64",
+)
 def test_retrace_system_sandbox_gdb_sandbox(
     workdir: pathlib.Path, module_cachedir: pathlib.Path, divide_by_zero_crash: str
 ) -> None:
@@ -256,6 +288,10 @@ def test_retrace_jammy_sandbox(
 
 
 @pytest.mark.skipif(not has_internet(), reason="online test")
+@pytest.mark.skipif(
+    impl.get_system_architecture() != "amd64",
+    reason="GDB sandbox only available on amd64",
+)
 def test_retrace_jammy_sandbox_gdb_sandbox(
     workdir: pathlib.Path, module_cachedir: pathlib.Path, sandbox_config: pathlib.Path
 ) -> None:
--- a/tests/system/test_apport_retrace.py
+++ b/tests/system/test_apport_retrace.py
@@ -142,17 +142,34 @@ def _assert_is_retraced(report: Report) -> None:
 
 
 def _assert_divide_by_zero_retrace(report: Report) -> None:
-    stack_top = "divide_by_zero () at /usr/src/chaos-marmosets"
-    if report["DistroRelease"] == "Ubuntu 22.04":
-        stack_top = "divide_by_zero () at ./divide-by-zero.c"
+    # The path could be relative to the source code (with or without
+    # a leading "./" or it can be absolute with /usr/src/<deb-source>-<deb-version>
+
+    # noble: at /usr/src/chaos-marmosets-0.2.0-1/divide-by-zero.c
+    # noble armhf: at divide-by-zero.c
+    # jammy: at ./divide-by-zero.c
+
+    # Actual frame info format depends on arch, compiler, opt-level...
+    # If the failing instruction is *not* the first instruction for the
+    # line there's an addition pointer in the line.
+    frame_regex = re.compile(
+        r"^#0  (0x[0-9a-f]+ in )?divide_by_zero \(\)"
+        r" at (/usr/src/chaos-marmosets-[^/]+/|\./)?divide-by-zero.c:[0-9]+$",
+        flags=re.M,
+    )
+    stack_regex = re.compile(
+        r"^divide_by_zero \(\)"
+        r" at (/usr/src/chaos-marmosets-[^/]+/|\./)?divide-by-zero.c:[0-9]+$",
+        flags=re.M,
+    )
     assert "divide_by_zero" in report["Disassembly"]
     # Expect RIP point to divide_by_zero
     assert "divide_by_zero" in report["Registers"]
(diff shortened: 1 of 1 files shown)


tests/system/test_apport_retrace.py
```diff
@@ -195,6 +205,38 @@ def test_retrace_system_sandbox(
 
 
 @pytest.mark.skipif(not has_internet(), reason="online test")
+@pytest.mark.skipif(
+    impl.get_system_architecture() == "amd64",
+    reason="GDB sandbox only available on amd64",
+)
+def test_retrace_system_sandbox_gdb_sandbox_nonamd64(
```
bdrung, line 212: This test is not executed in the CI. Could we convert it into a unit test and mock the architecture?

schopin-pro: I'd rather not. Basically, you want to make the test less meaningful for the convenience of running it on the GH CI. I'd much rather have this code path exercised for *real* in the Ubuntu CI, if less frequently.

tests/system/test_apport_retrace.py
```diff
@@ -195,6 +205,38 @@ def test_retrace_system_sandbox(
...
+    workdir: pathlib.Path, module_cachedir: pathlib.Path, divide_by_zero_crash: str
+) -> None:
+    """Retrace a divide-by-zero crash in a system sandbox with a GDB sandbox."""
```
bdrung, line 215: todo: It has the same description that the copy-paste origin.

tests/system/test_apport_retrace.py
```diff
@@ -195,6 +205,38 @@ def test_retrace_system_sandbox(
...
+        "-o",
+        str(retraced_report_filename),
+        "--sandbox",
+        "system",
+        "--gdb-sandbox",
+        "--cache",
+        str(module_cachedir),
+        divide_by_zero_crash,
+    ]
+    ret = subprocess.run(cmd, check=False, env=env, capture_output=True)
```
bdrung, line 230: suggestion: you could use `text=True` no avoid the `.decode()` later.

tests/system/test_apport_retrace.py
```diff
@@ -195,6 +205,38 @@ def test_retrace_system_sandbox(
 
 
 @pytest.mark.skipif(not has_internet(), reason="online test")
+@pytest.mark.skipif(
+    impl.get_system_architecture() == "amd64",
+    reason="GDB sandbox only available on amd64",
```
bdrung, line 210: nitpick: this reason is confusing, but i fail to come up with a clearer wording

tests/system/test_apport_retrace.py
```diff
@@ -136,17 +137,26 @@ def _assert_is_retraced(report: Report) -> None:
 
 
 def _assert_divide_by_zero_retrace(report: Report) -> None:
+    # The path could be relative to the so
+
+    # noble: at /usr/src/chaos-marmosets-0.2.0-1/divide-by-zero.c
+    # noble armhf: at divide-by-zero.c
+    # jammy: at ./divide-by-zero.c
```
bdrung, line 144: There seems to be missing something from the first comment line. In the code snippet that I have from our coding session I have following instead:
```
    # The path could be relative to the source code (with or without
    # a leading "./") or it can be absolute with
    # /usr/src/<deb-source>-<deb-version>/
```

tests/system/test_apport_retrace.py
```diff
@@ -136,17 +137,26 @@ def _assert_is_retraced(report: Report) -> None:
...
     stack_top = "divide_by_zero () at /usr/src/chaos-marmosets"
     if report["DistroRelease"] == "Ubuntu 22.04":
         stack_top = "divide_by_zero () at ./divide-by-zero.c"
```
bdrung, line 147: suggestion: Instead of those, we can accept all variants in a regular expression:
```
    frame_regex = re.compile(
        r"^#0  (0x[0-9a-b]+ in )? divide_by_zero \(\)"
        r" at (/usr/src/chaos-marmosets-[^/]+/|./)?divide-by-zero.c:[0-9]+$",
        flags=re.M,
    )
    stack_regex = re.compile(
        r"^divide_by_zero \(\)"
        r" at (/usr/src/chaos-marmosets-[^/]+/|./)?divide-by-zero.c:[0-9]+$",
        flags=re.M,
    )
```

tests/system/test_apport_retrace.py
```diff
@@ -136,17 +137,26 @@ def _assert_is_retraced(report: Report) -> None:
...
+    frame_regex = re.compile(rf"#0  (0x[0-9a-f]+ in )?{re.escape(stack_top)}")
     assert "divide_by_zero" in report["Disassembly"]
     # Expect RIP point to divide_by_zero
     assert "divide_by_zero" in report["Registers"]
-    assert f"#0  {stack_top}" in report["Stacktrace"]
-    assert f"#0  {stack_top}" in report["StacktraceSource"]
+    assert frame_regex.match(report["Stacktrace"])
+    assert frame_regex.match(report["StacktraceSource"])
     assert "42 / zero" in report["StacktraceSource"]
     assert stack_top in report["StacktraceTop"]
```
bdrung, line 158: The comment above: `# noble armhf: at divide-by-zero.c`. So this assertion will probably fail on armhf.

bdrung (changes requested):

tests/system/test_apport_retrace.py
bdrung: Okay.

tests/system/test_apport_retrace.py
schopin-pro: So do I.

tests/system/test_apport_retrace.py
schopin-pro: Right, I think you wrote that comment, because I have no memory of it (I think I don't pay much attention to comments when reviewing my commits. I should work on that.)

tests/system/test_apport_retrace.py
schopin-pro: That should be handled by your suggestion, thanks!

tests/system/test_apport_retrace.py
```diff
@@ -136,17 +137,34 @@ def _assert_is_retraced(report: Report) -> None:
...
+
+    # noble: at /usr/src/chaos-marmosets-0.2.0-1/divide-by-zero.c
+    # noble armhf: at divide-by-zero.c
+    # jammy: at ./divide-by-zero.c
+
+    # Actual frame info format depends on arch, compiler, opt-level...
+    # If the failing instruction is *not* the first instruction for the
+    # line there's an addition pointer in the line.
+    frame_regex = re.compile(
+        r"^#0  (0x[0-9a-b]+ in )? divide_by_zero \(\)"
```
bdrung, line 151: my bad: I broke the hex-value expression.
```suggestion
        r"^#0  (0x[0-9a-f]+ in )? divide_by_zero \(\)"
```

tests/system/test_apport_retrace.py
```diff
@@ -195,6 +213,38 @@ def test_retrace_system_sandbox(
 
 
 @pytest.mark.skipif(not has_internet(), reason="online test")
+@pytest.mark.skipif(
+    impl.get_system_architecture() == "amd64",
+    reason="GDB sandbox only available on amd64",
```
bdrung, line 218:
```suggestion
    reason="GDB sandbox is available on amd64",
```

tests/system/test_apport_retrace.py
schopin-pro: Boooo, bad bdrung ;).
I also spotted another small mistake (a non-escaped `.`) a little further down.

schopin-pro: Also, I know where that's from. That's actually my original mistake that I fixed back when I created the commits, but that was after our pairing session so your local copy never got the fix.

bdrung: Since frame_regex is a longer version of stack_regex, you might want to deduplicate the code.

schopin-pro: I thought about it already, but the easy path was thwarted by the `^` and then the dedup code was beginning to be a bit too ugly, so I decided to scrap it.

bdrung (approved):
The changes look good.

I'll verify that these changes plus other commits will make the test succeed on all archs via apport 2.29.0+git20240719b-0ubuntu1~ppa1 in https://launchpad.net/~bdrung/+archive/ubuntu/apport

Once those autopkgtests are green, I'll merged this PR.

bdrung:
After fixing the memory consumption of apport-retrace, these tests succeed on arm64, armhf, ppc64el. Only s390x fails to retrace (but that looks like a separate issue). So we are good to go.

PR #360, opened 2024-07-09
apport-retrace: exit without 0 on error

bdrung:
Exit `apport-retrace` with return code 2 in case the report is malformed and with return code 3 on known errors. Crashes of `apport-retrace` will exit with return code 1.

### DIFF
--- a/bin/apport-retrace
+++ b/bin/apport-retrace
@@ -353,7 +353,7 @@ def load_report(
         except AssertionError as error:
             if "apport format data" in str(error):
                 apport.error("Broken report: %s", str(error))
-                raise _LoadReportException(0) from None
+                raise _LoadReportException(2) from None
             raise
         except (
             MemoryError,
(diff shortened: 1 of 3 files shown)


bin/apport-retrace
```diff
@@ -349,7 +349,7 @@ def load_report(
         except AssertionError as error:
             if "apport format data" in str(error):
                 apport.error("Broken report: %s", str(error))
-                sys.exit(0)
+                sys.exit(2)
```
schopin-pro, line 352: issue: it's a shame to have main() return an int but still have a hard exit() call in `load_report()`. How about throwing an exception instead, in proper Python fashion?

bin/apport-retrace
```diff
@@ -391,15 +391,15 @@ Thank you for your understanding, and sorry for the inconvenience!
 
 
 # pylint: disable-next=missing-function-docstring
-def main():
+def main(argv):
```
schopin-pro, line 394: praise: Nice, makes it easier to test!

schopin-pro (changes requested):
Very  nice, except for the load_report() bit where we're missing the opportunity to remove a raw exit() call.

bin/apport-retrace
bdrung: Okay. Introduced `_LoadReportException`.

schopin-pro (approved):

PR #367, opened 2024-07-16
launchpad: don't systematically order retrace on private reports (FR-7945)

schopin-pro:
Just because a report is marked as LaunchpadPrivate and has an actual
architecture attached to it doesn't mean it should be retraced. A
contrario, any report that is designated as LaunchpadPrivate should be
marked as private, even if get_os_info() was never called on it.

Fixes https://bugs.launchpad.net/ubuntu/+source/apport/+bug/2068933

The review will be much easier if done commit-wise.

### DIFF
--- a/apport/crashdb_impl/launchpad.py
+++ b/apport/crashdb_impl/launchpad.py
@@ -27,6 +27,7 @@ import tempfile
 import time
 import urllib.parse
 import urllib.request
+from typing import IO
 
 try:
     from httplib2 import FailedToDecompressContent
@@ -1063,25 +1064,22 @@ class CrashDatabase(apport.crashdb.CrashDatabase):
         ]:
             bug.subscribe(person=person)
 
-    def _generate_upload_blob(self, report):
-        """Generate a multipart/MIME temporary file for uploading.
-
-        You have to close the returned file object after you are done with it.
-        """
+    def _generate_upload_headers(self, report: apport.Report) -> dict[str, str]:
+        """Generate the headers for the multipart/MIME temporary file."""
         # set reprocessing tags
         hdr = {"Tags": f"apport-{report['ProblemType'].lower()}"}
-        a = report.get("PackageArchitecture")
-        if not a or a == "all":
-            a = report.get("Architecture")
-        if a:
-            hdr["Tags"] += f" {a}"
+        arch = report.get("PackageArchitecture")
+        if not arch or arch == "all":
+            arch = report.get("Architecture")
+        if arch:
+            hdr["Tags"] += f" {arch}"
         if "Tags" in report:
             hdr["Tags"] += f" {self._filter_tag_names(report['Tags'])}"
 
         # privacy/retracing for distro reports
         # FIXME: ugly hack until LP has a real crash db
         if "DistroRelease" in report:
-            if a and (
+            if arch and (
                 "VmCore" in report
                 or "CoreDump" in report
                 or "LaunchpadPrivate" in report
@@ -1091,11 +1089,12 @@ class CrashDatabase(apport.crashdb.CrashDatabase):
                     "LaunchpadSubscribe",
                     self.options.get("initial_subscriber", "apport"),
                 )
-                hdr["Tags"] += f" need-{a}-retrace"
+                hdr["Tags"] += f" need-{arch}-retrace"
             elif "Traceback" in report:
                 hdr["Private"] = "yes"
                 hdr["Subscribers"] = "apport"
                 hdr["Tags"] += " need-duplicate-check"
+
         if "DuplicateSignature" in report and "need-duplicate-check" not in hdr["Tags"]:
             hdr["Tags"] += " need-duplicate-check"
 
@@ -1103,7 +1102,14 @@ class CrashDatabase(apport.crashdb.CrashDatabase):
         # reference until the link is shown in Launchpad's UI
         if "CheckboxSubmission" in report:
             hdr["HWDB-Submission"] = report["CheckboxSubmission"]
+        return hdr
 
+    def _generate_upload_blob(self, report: apport.Report) -> IO[bytes]:
+        """Generate a multipart/MIME temporary file for uploading.
+
+        You have to close the returned file object after you are done with it.
+        """
+        hdr = self._generate_upload_headers(report)
         # order in which keys should appear in the temporary file
         order = [
             "ProblemType",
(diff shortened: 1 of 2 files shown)


apport/crashdb_impl/launchpad.py
```diff
@@ -1146,6 +1134,28 @@ def _filter_tag_names(tags):
...
 
+    def _enforce_privacy_for_distro_reports(
+        self, report: apport.report.Report, hdr: dict[str, str], arch: str | None
+    ) -> None:
+        # privacy/retracing for distro reports
+        # FIXME: ugly hack until LP has a real crash db
+        if "DistroRelease" not in report:
+            return
+
+        if "LaunchpadPrivate" in report:
```
bdrung, line 1145: question: Shouldn't the LaunchpadPrivate check done before the DistroRelease check?

apport/crashdb_impl/launchpad.py
```diff
@@ -1063,47 +1063,35 @@ def _subscribe_triaging_team(self, bug, report):
         ]:
             bug.subscribe(person=person)
 
-    def _generate_upload_blob(self, report):
-        """Generate a multipart/MIME temporary file for uploading.
-
-        You have to close the returned file object after you are done with it.
-        """
+    def _generate_upload_headers(self, report: apport.report.Report) -> dict[str, str]:
```
bdrung, line 1066: praise: Splitting out the header generation is nice.

apport/crashdb_impl/launchpad.py
```diff
@@ -1146,6 +1134,28 @@ def _filter_tag_names(tags):
 
         return res
 
+    def _enforce_privacy_for_distro_reports(
```
bdrung, line 1137: This method name is misleading. It does not tell me that it also sets Subscribers and Tags. idea: maybe split this function into two: one just for checking for privacy and the other(s) for the retrace/duplicate checks.

bdrung (changes requested):

tests/unit/test_launchpad.py
```diff
@@ -0,0 +1,93 @@
...
+    report["Package"] = "python-goo 3epsilon1"
+    report["SourcePackage"] = "pygoo"
+    report["PackageArchitecture"] = "all"
+    report["DistroRelease"] = "Ubuntu 24.04"
+    report["ExecutablePath"] = "/usr/bin/pygoo"
+    report[
+        "Traceback"
+    ] = """Traceback (most recent call last):
+File "test.py", line 7, in <module>
+print(_f(5))
```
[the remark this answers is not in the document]
schopin-pro: Not sure, I lifted this from another test file :D

apport/crashdb_impl/launchpad.py
schopin-pro: Maybe? I don't really want to change the logic too much here unless someone explicitly starts to complain.

bdrung: Do you want me to explicitly complain?

schopin-pro: Not unless you actually *have* a problem. We don't fully understand the scope of this, what the `FIXME` was all about, etc. Let's avoid gratuitous behaviour changes, shall we?

apport/crashdb_impl/launchpad.py
schopin-pro: Agreed on the misleading part, but I'd rather fix it by bringing that code back up the stack rather than splitting it further. The only reason this function exists in the first place is that I started writing the tests for it before I decided to go wider, and splitting it is just a good way to introduce unintended changes.

WDYT?

bdrung: I am fine with moving it back to `_generate_upload_headers`

apport/crashdb_impl/launchpad.py
bdrung: crashdb-conf.md says:
> Crash reports are always filed as private Launchpad bug. Bug reports are public by default, but a package hook can change this by adding a `LaunchpadPrivate` report field
It does not say that `LaunchpadPrivate` requires `DistroRelease` to be set in the report for it to work.

schopin-pro: This could easily be shown as an example of scope creep. I'm here to fix the need-$arch-retrace thing. The fact that the doc doesn't match the current behaviour (which has been the same for a decade) is fairly unrelated, and changing it can have consequences that I for one don't fully understand.

Please put down the stick and let's bury that horse.

tests/unit/test_launchpad.py
```diff
@@ -0,0 +1,93 @@
...
+def native_crash() -> Report:
+    """Generate a report that looks like a native binary crash"""
+    report = Report("Crash")
+    report["Signal"] = "6"
+    report["Package"] = "bash"
+    report["SourcePackage"] = "bash"
+    report["DistroRelease"] = "Ubuntu 24.04"
+    report["PackageArchitecture"] = "i386"
+    report["Architecture"] = "amd64"
+    report["SignalName"] = "SIGABRT"
```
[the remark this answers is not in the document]
schopin-pro: I won't sort this because some related fields would be far away from each other, but yeah, Signals should be next to each other.

tests/unit/test_launchpad.py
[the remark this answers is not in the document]
schopin-pro: Fixed.

apport/crashdb_impl/launchpad.py
bdrung: I was mislead and though it was about `LaunchpadPrivate`. So fixing that can go in a different merge request.

bdrung (approved):
Approved. Will merged once CI is green.

tests/unit/test_crashdb_launchpad.py
```diff
@@ -0,0 +1,93 @@
...
+    report["Package"] = "apport"
+    report["SourcePackage"] = "apport"
+    report["PackageArchitecture"] = "all"
+    report["Architecture"] = "amd64"
+    report["DistroRelease"] = "Ubuntu 24.04"
+    report["LaunchpadPrivate"] = "yes"
+    headers = crashdb._generate_upload_headers(report)
+
+    assert headers.get("Private") == "yes"
+    assert not re.match(r"need-[a-z0-9]-retrace", headers.get("Tags", ""))
```
bdrung, line 93: Have you tested that test to fail before the fix is applied? I tried that and found two issues here:
```suggestion
    assert not re.search(r"need-[a-z0-9]+-retrace", headers.get("Tags", ""))
```

bdrung (changes requested):

tests/unit/test_crashdb_launchpad.py
schopin-pro: Oh god I really hate this particular Python API. Trips me up every time.

schopin-pro: And no, I apparently didn't run the test *before* fixing the code. Good catch!

PR #368, opened 2024-07-17
test: rename pidof to pids_of

bdrung:
Rename the function `pidof` to `pids_of` to reflect that it will return a set of process IDs. This was requested in https://github.com/canonical/apport/pull/358#discussion_r1679388131.

### DIFF
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -52,8 +52,8 @@ def import_module_from_file(path: pathlib.Path) -> Any:
     return module
 
 
-def pidof(program: str) -> set[int]:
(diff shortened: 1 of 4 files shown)


schopin-pro (changes requested):
Actually the most important part of my suggestion wasn't the snake_case (not to be confused with CamelCase) but rather the pluralisation (pid -> pids) ;)

bdrung:
Corrected.

That would make a great T-shirt:
> Use camel_case instead of SnakeCase!

schopin-pro (approved):

PR #375, opened 2024-07-19
CI: only run the linters on a single configuration

schopin-pro:
Having the same linters run on multiple configurations is IMHO a waste of resources, for gains that have been thus far theoretical, and with actual costs due to discrepancies in the analysis for the various versions.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,14 +16,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      matrix:
-        container:
-          - debian:testing-slim
-          - ubuntu:jammy
-          - ubuntu:noble
-          - ubuntu:oracular
     container:
-      image: ${{ matrix.container }}
+      image: ubuntu:oracular
     steps:
       - uses: actions/checkout@v4
       - name: Install dependencies
@@ -33,19 +27,8 @@ jobs:
           black gir1.2-gtk-3.0 gir1.2-wnck-3.0 isort mypy pycodestyle
           pydocstyle pyflakes3 pylint python3 python3-apt python3-dbus
           python3-distutils-extra python3-gi python3-launchpadlib
-          python3-psutil python3-pytest python3-rpm python3-typeshed
-          python3-yaml python3-systemd
-      - name: Install python3-pyqt5 on non-jammy (to avoid false positives)
-        run: >
-          . /etc/os-release &&
-          if test "${VERSION_CODENAME-}" != jammy; then
-            apt-get install --no-install-recommends --yes python3-pyqt5;
-          fi
-      # python3-zstandard is not available on ubuntu:jammy
-      - name: Install optional dependencies
-        run: >
-          apt-get install --no-install-recommends --yes python3-zstandard
-          || true
+          python3-psutil python3-pyqt5 python3-pytest python3-rpm
+          python3-typeshed python3-yaml python3-systemd python3-zstandard
       - name: Run linter tests
         run: tests/run-linters
 


.github/workflows/ci.yaml
```diff
@@ -13,17 +13,9 @@ env:
 
 jobs:
   linter:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu:oracular
     strategy:
       fail-fast: false
-      matrix:
-        container:
-          - debian:testing-slim
```
bdrung, line 21: Running the linter on Debian testing had the benefit of detecting issues with newer linters earlier.

.github/workflows/ci.yaml
```diff
@@ -34,7 +26,7 @@ jobs:
           pydocstyle pyflakes3 pylint python3 python3-apt python3-dbus
           python3-distutils-extra python3-gi python3-launchpadlib
           python3-psutil python3-pytest python3-rpm python3-typeshed
-          python3-yaml python3-systemd
+          python3-yaml python3-systemd python3-zstandard
       - name: Install python3-pyqt5 on non-jammy (to avoid false positives)
```
bdrung, line 30: That one can be merged with the step before.

bdrung (commented):
I would prefer to only drop jammy (and maybe noble).

bdrung:
Here is an example, where the linter run on jammy caught an issue:

```
Running mypy...
apport/packaging_impl/apt_dpkg.py:175: error: "Connection" has no attribute "autocommit"
```

Log: https://github.com/bdrung/apport/actions/runs/10085379790/job/27886060441

.github/workflows/ci.yaml
schopin-pro: NACK. The point of this PR is to have consistent linter diagnostics. If we're diligent in using the devel series of Ubuntu (which is our primary target) we'll be 4 months behind at the most, which seems just fine to me.

.github/workflows/ci.yaml
```diff
@@ -13,17 +13,9 @@ env:
...
     strategy:
       fail-fast: false
-      matrix:
-        container:
-          - debian:testing-slim
-          - ubuntu:jammy
-          - ubuntu:noble
-          - ubuntu:oracular
-    container:
-      image: ${{ matrix.container }}
```
bdrung, line 26: You need to set the container here (there is no `runs-on: ubuntu:oracular`).

schopin-pro:
> Here is an example, where the linter run on jammy caught an issue:
> 
> ```
> Running mypy...
> apport/packaging_impl/apt_dpkg.py:175: error: "Connection" has no attribute "autocommit"
> ```


And my point was never that this doesn't happen. My point is that this is actually *harmful*. Jammy is NOT OUR TARGET! Our target is the `devel` series of Ubuntu, and I guess it's nice if we support the last LTS too, just in case, but that shouldn't even be blocking.

Because here one linter caught an actual bug (I think, didn't check), but it also happens that another linter changes its rules over time, making us bend over backwards to try to appease them. If I take the time to actually run the linters on *each commit* of my PRs before pushing, if the CI gets mad at me it better be for an actual bug rather than pylint having relaxed its rules on single-letter variables over time. Linters are supposed to make my life **easier**.

.github/workflows/ci.yaml
```diff
@@ -33,19 +27,8 @@ jobs:
...
-          . /etc/os-release &&
-          if test "${VERSION_CODENAME-}" != jammy; then
-            apt-get install --no-install-recommends --yes python3-pyqt5;
-          fi
-      # python3-zstandard is not available on ubuntu:jammy
-      - name: Install optional dependencies
-        run: >
-          apt-get install --no-install-recommends --yes python3-zstandard
-          || true
+          python3-psutil python3-pytest  python3-qt5 python3-rpm
```
bdrung, line 30:
```suggestion
          python3-psutil python3-pytest python3-qt5 python3-rpm
```

bdrung (approved):

.github/workflows/ci.yaml
```diff
@@ -33,19 +27,8 @@ jobs:
...
-          . /etc/os-release &&
-          if test "${VERSION_CODENAME-}" != jammy; then
-            apt-get install --no-install-recommends --yes python3-pyqt5;
-          fi
-      # python3-zstandard is not available on ubuntu:jammy
-      - name: Install optional dependencies
-        run: >
-          apt-get install --no-install-recommends --yes python3-zstandard
-          || true
+          python3-psutil python3-pytest python3-qt5 python3-rpm
```
bdrung, line 30: python3-qt5 -> python3-pyqt5
```suggestion
          python3-psutil python3-pyqt5 python3-pytest python3-rpm
```

PR #378, opened 2024-07-24
apt_dpkg: speed up Contents paths filtering with regular expression and fix parsing paths with spaces

bdrung:
In case no Contents cache exists, `apport-retrace` spends most execution time in `_update_given_file2pkg_mapping` and bytes splitting. Speed up Contents paths filtering with one regular expression. To prevent regressions add a test case that test all previous cases.

Result from parsing the four jammy Contents files on a Ryzen 7 5700G machine: It took around 60 seconds before and 41 seconds afterwards.

The regular expression filters only unwanted directories and is less strict than the previous solution. So they will include multi-arch paths that we missed before. The number of additional paths added to the database are neglectable:

* jammy: 7,861,981 -> 7,903,339 (0.48 %)
* noble: 4,201,611 -> 4,266,551 (1.55 %)

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -1576,13 +1576,13 @@ class __AptDpkgPackageInfo(PackageInfo):
         file2pkg: dict[bytes, bytes], contents_filename: str, dist: str
     ) -> None:
         with gzip.open(contents_filename, "rb") as contents:
-            line_num = 0
-            for line in contents:
-                line_num += 1
+            if dist in {"trusty", "xenial"}:
                 # the first 32 lines are descriptive only for these
                 # releases
-                if dist in {"trusty", "xenial"} and line_num < 33:
-                    continue
+                for _ in range(32):
+                    next(contents)
+
+            for line in contents:
                 path = line.split()[0]
                 if path.split(b"/")[0] == b"usr":
                     if path.split(b"/")[1] not in (
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -1575,6 +1575,10 @@ class __AptDpkgPackageInfo(PackageInfo):
     def _update_given_file2pkg_mapping(
         file2pkg: dict[bytes, bytes], contents_filename: str, dist: str
     ) -> None:
+        path_exclude_pattern = re.compile(
+            rb"^:|(boot|var|usr/(include|src|[^/]+/include"
+            rb"|share/(doc|gocode|help|icons|locale|man|texlive)))/"
+        )
         with gzip.open(contents_filename, "rb") as contents:
             if dist in {"trusty", "xenial"}:
                 # the first 32 lines are descriptive only for these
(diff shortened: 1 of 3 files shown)


apport/packaging_impl/apt_dpkg.py
```diff
@@ -1575,42 +1575,22 @@ def _get_contents_file(self, map_cachedir: str, dist: str, arch: str) -> str | N
     def _update_given_file2pkg_mapping(
         file2pkg: dict[bytes, bytes], contents_filename: str, dist: str
     ) -> None:
+        path_exclude_pattern = re.compile(
+            rb"^:|(boot|var|usr/(include|src|[^/]+/include"
```
schopin-pro, line 1579: question: what is this colon doing here?

apport/packaging_impl/apt_dpkg.py
```diff
@@ -1576,13 +1576,13 @@ def _update_given_file2pkg_mapping(
         file2pkg: dict[bytes, bytes], contents_filename: str, dist: str
     ) -> None:
         with gzip.open(contents_filename, "rb") as contents:
-            line_num = 0
-            for line in contents:
-                line_num += 1
+            if dist in {"trusty", "xenial"}:
```
schopin-pro, line 1583: praise: that's a nice gain for such a small change :)

apport/packaging_impl/apt_dpkg.py
```diff
@@ -1576,13 +1576,13 @@ def _update_given_file2pkg_mapping(
...
                 # the first 32 lines are descriptive only for these
                 # releases
-                if dist in {"trusty", "xenial"} and line_num < 33:
-                    continue
+                for _ in range(32):
+                    next(contents)
```
schopin-pro, line 1587: suggestion: Maybe use dedicated operators from itertools, e.g. islice() or dropwhile()?

I tend to think that iterating manually an iterator is a sign that we're reinventing a wheel (but maybe I was just spoiled with Rust rich iterator story)

schopin-pro (approved):
LGTM! I do have a small question, but feel free to go ahead and merge.

apport/packaging_impl/apt_dpkg.py
bdrung: That one is no exclude `:sexsend:sexget:` from the xenial Contents (see `test_contents_skip_xenial_header`).

apport/packaging_impl/apt_dpkg.py
bdrung: TIL `itertools.islice`. Those two lines could be changed to:
```
                contents = itertools.islice(contents, 32, None)
```
but then mypy will complain about the changed type:
```
apport/packaging_impl/apt_dpkg.py:1587: error: Incompatible types in assignment (expression has type "islice[bytes]", variable has type "GzipFile")  [assignment]
```
Suggestions here?

bdrung: Okay, changing the with statement would work:
```python3
        with gzip.open(
            contents_filename, "rb"
        ) as contents:  # type: gzip.GzipFile | itertools.islice[bytes]
```
Do you still prefer that change?

bdrung:
```patch
diff --git a/apport/packaging_impl/apt_dpkg.py b/apport/packaging_impl/apt_dpkg.py
index 7ff23eeb2..83326dc03 100644
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -29,6 +29,7 @@ import glob
 import gzip
 import hashlib
 import http.client
+import itertools
 import json
 import logging
 import os
@@ -1579,12 +1580,13 @@ class __AptDpkgPackageInfo(PackageInfo):
             rb"^:|(boot|var|usr/(include|src|[^/]+/include"
             rb"|share/(doc|gocode|help|icons|locale|man|texlive)))/"
         )
-        with gzip.open(contents_filename, "rb") as contents:
+        with gzip.open(
+            contents_filename, "rb"
+        ) as contents:  # type: gzip.GzipFile | itertools.islice[bytes]
             if dist in {"trusty", "xenial"}:
                 # the first 32 lines are descriptive only for these
                 # releases
-                for _ in range(32):
-                    next(contents)
+                contents = itertools.islice(contents, 32, None)
 
             for line in contents:
                 if path_exclude_pattern.match(line):
```

schopin-pro: Let leave the manual loop. That goes to show that bolting static typing on a strongly-type dynamic language is far from perfect (and would support relaxing our current requirements)

PR #379, opened 2024-07-30
"Fix" the apport-retrace tests on s390x

schopin-pro:
This contains a small fix as well as a workaround for the underlying issue (which is that GDB can't cope with divide-by-zero on s390x)

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -705,6 +705,18 @@ class __AptDpkgPackageInfo(PackageInfo):
         assert arch
         return arch
 
+    @staticmethod
+    @functools.cache
+    def get_native_multiarch_triplet() -> str:
+        """Return the multiarch triplet for the system architecture"""
+        dpkg = subprocess.run(
+            ["dpkg-architecture", "-qDEB_HOST_MULTIARCH"],
+            check=True,
+            text=True,
+            stdout=subprocess.PIPE,
+        )
+        return dpkg.stdout.strip()
+
     def get_library_paths(self) -> str:
         """Return a list of default library search paths.
 
@@ -712,14 +724,7 @@ class __AptDpkgPackageInfo(PackageInfo):
         $LD_LIBRARY_PATH. This needs to take any multiarch directories into
         account.
         """
-        dpkg = subprocess.run(
-            ["dpkg-architecture", "-qDEB_HOST_MULTIARCH"],
-            check=True,
-            stdout=subprocess.PIPE,
-        )
-        multiarch_triple = dpkg.stdout.decode().strip()
-
-        return f"/lib/{multiarch_triple}:/lib"
+        return f"/lib/{self.get_native_multiarch_triplet()}:/lib"
 
     def set_mirror(self, url: str) -> None:
         """Explicitly set a distribution mirror URL for operations that need to
--- a/apport/report.py
+++ b/apport/report.py
@@ -1887,7 +1887,7 @@ class Report(problem_report.ProblemReport):
         environ: dict[str, str] = {}
 
         if sandbox:
-            native_multiarch = "x86_64-linux-gnu"
+            native_multiarch = packaging.get_native_multiarch_triplet()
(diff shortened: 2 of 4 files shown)


apport/packaging_impl/apt_dpkg.py
```diff
@@ -700,21 +700,25 @@ def get_system_architecture():
         assert arch
         return arch
 
+    @staticmethod
+    @functools.cache
+    def get_multiarch_triplet():
+        """Return the multiarch triplet for the system architecture"""
+        dpkg = subprocess.run(
+            ["dpkg-architecture", "-qDEB_HOST_MULTIARCH"],
```
bdrung, line 708: `dpkg-architecture` is shipped by `dpkg-dev` which we cannot rely on being present.

apport/report.py
```diff
@@ -1887,7 +1887,7 @@ def gdb_command(
         environ: dict[str, str] = {}
 
         if sandbox:
-            native_multiarch = "x86_64-linux-gnu"
+            native_multiarch = packaging.get_multiarch_triplet()
```
bdrung, line 1890: Don't you want the triplet for `self["Architecture"]`?

tests/system/test_apport_retrace.py
```diff
@@ -197,6 +197,10 @@ def _assert_cache_has_content(
 
 
 @pytest.mark.skipif(not has_internet(), reason="online test")
+@pytest.mark.skipif(
+    impl.get_system_architecture() == "s390x",
+    reason="GDB has issues with divide-by-zero on s390x (LP: #2075204)",
+)
```
bdrung, line 203: question: Do we want to carry this patch in the upstream repository or only as patch in the Ubuntu package?

schopin-pro: I say we keep it, since we use Ubuntu as our development platform and I don't want this (very long) test to get in the way when investigating another s390x failure somewhere else, which I usually do from the upstream repo sooner rather than later.

apport/report.py
schopin-pro: The variable says "native_multiarch", that's exactly what you get there.

apport/packaging_impl/apt_dpkg.py
schopin-pro: For apport-retrace, we can, it's an explicit dependency.

bdrung: Oh, nice.

apport/packaging_impl/apt_dpkg.py
```diff
@@ -700,21 +700,25 @@ def get_system_architecture():
         assert arch
         return arch
 
+    @staticmethod
+    @functools.cache
+    def get_multiarch_triplet():
```
bdrung, line 705: suggestion: add type hint to it?

apport/packaging_impl/apt_dpkg.py
```diff
@@ -700,21 +700,25 @@ def get_system_architecture():
...
+        """Return the multiarch triplet for the system architecture"""
+        dpkg = subprocess.run(
+            ["dpkg-architecture", "-qDEB_HOST_MULTIARCH"],
+            check=True,
+            stdout=subprocess.PIPE,
+        )
+        return dpkg.stdout.decode().strip()
```
bdrung, line 712: suggestion: call `subprocess.run` with `text=True`. Then the `.decode()` is not needed.

apport/report.py
bdrung: Thanks for pointing that out. I missed that.

suggestion: add `native` to the `get_multiarch_triplet` method name for clarity.

bdrung: I am still not sure if the native multiarch should be used for `solib-search-path`, but that is out of scope for this merge request (and should be addressed in a later merge request if needed).

bdrung (commented):
suggestion: Can you reference Bug-Ubuntu: https://launchpad.net/bugs/2069815 in the first commit message?

bdrung (approved):

PR #382, opened 2024-08-02
recent-syslog: read stdout after process completion

Chris-Peterson444:
The recent_syslog integration tests were failing autopkgtest due to getting an empty string for the syslog. This was caused by the syslog command finishing before the python code entered the poll loop in `__filter_re_process` and also not checking that the buffer was empty on completion (LP:[#2073935](https://pad.lv/2073935)).

Originally this bug was reported against s390x since it reproduced often there, but there's nothing architecture specific about the issue (other than the process scheduling it seems). For completeness, you can reproduce the output with the following code:

```python
from subprocess import Popen, PIPE
import time

with Popen(["echo", "content"], stdout=PIPE) as p:
    time.sleep(1)
    pre_complete_content = ""
    post_complete_content = ""
    while p.poll() is None:
        for line in p.stdout:
            pre_complete_content += line.decode("UTF-8")
    p.wait()
    for line in p.stdout:
        post_complete_content += line.decode("UTF-8")

print(f"{pre_complete_content=}")
print(f"{post_complete_content=}")
```
```
cpete@atlas:$ python3 test-fail.py
pre_complete_content=''
post_complete_content='content\n'
```

### DIFF
--- a/tests/unit/test_hookutils.py
+++ b/tests/unit/test_hookutils.py
@@ -6,7 +6,7 @@ import re
 import subprocess
 import unittest
 import unittest.mock
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, Mock
 
 import apport.hookutils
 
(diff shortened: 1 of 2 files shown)


tests/unit/test_hookutils.py
```diff
@@ -186,3 +186,35 @@ def test_deprecated_upstart_functions(self) -> None:
             apport.hookutils.attach_upstart_logs({}, "apport")
         with self.assertWarns(PendingDeprecationWarning):
             apport.hookutils.attach_upstart_overrides({}, "nonexisting")
+
+    def test_recent_syslog_race_condition(self) -> None:
+        """Test recent_syslog is not racing process completion."""
+        # LP: #2073935
+
+        # Test read buffer after process completion only
```
bdrung, line 194: These comment could be moved into the docstring.

Chris-Peterson444: Done!

Chris-Peterson444:
Apologies for all the noise. I think this is ready now. As discussed elsewhere, this is likely also the fix for [LP: #2073934](https://bugs.launchpad.net/ubuntu/+source/apport/+bug/2073934).

tests/unit/test_hookutils.py
```diff
@@ -186,3 +186,45 @@ def test_deprecated_upstart_functions(self) -> None:
...
+        """
+        with unittest.mock.patch("apport.hookutils.subprocess.Popen") as popen:
+            process = unittest.mock.Mock()
+            popen.return_value.__enter__.return_value = process
+
+            process.poll = unittest.mock.Mock(return_value=0)
+            process.returncode = 0
+            buffer_mock = unittest.mock.MagicMock()
+            buffer_mock.__iter__.side_effect = [iter([b"content"])]
+            process.stdout = buffer_mock
```
bdrung, line 204: idea: Maybe that is more readable?
```suggestion
            process.stdout = unittest.mock.MagicMock()
            process.stdout.__iter__.side_effect = [iter([b"content"])]
```

tests/unit/test_hookutils.py
```diff
@@ -186,3 +186,45 @@ def test_deprecated_upstart_functions(self) -> None:
...
         with self.assertWarns(PendingDeprecationWarning):
             apport.hookutils.attach_upstart_overrides({}, "nonexisting")
+
+    def test_recent_syslog_race_condition(self) -> None:
+        """Test recent_syslog is not racing process completion.
+
+        LP: #2073935
+        Test read buffer after process completion.
+        """
+        with unittest.mock.patch("apport.hookutils.subprocess.Popen") as popen:
```
bdrung, line 196: suggestion: normally we name mock objects `*_mock`, e.g. `popen_mock`.

tests/unit/test_hookutils.py
```diff
@@ -186,3 +186,45 @@ def test_deprecated_upstart_functions(self) -> None:
...
+            process = unittest.mock.Mock()
+            popen.return_value.__enter__.return_value = process
+
+            process.poll = unittest.mock.Mock(return_value=0)
+            process.returncode = 0
+            buffer_mock = unittest.mock.MagicMock()
```
bdrung, line 202:
```suggestion
            buffer_mock = MagicMock()
```

Chris-Peterson444:
Thanks for the suggestions @bdrung. I incorporated all of your suggestions and slightly modified the docstrings as well.

bdrung (approved):

PR #383, opened 2024-08-06
test: wait_for_gdb_sleeping_child_process

Chris-Peterson444:
- `wait_for_gdb_child_process` is used exclusively with the sleep command. Rename to `wait_for_gdb_sleeping_child_process` to make the context of function clearer.

- Modify the wait_for_gdb_sleeping_child_process utility function to make sure to wait until the test process is actually sleeping. The expected behavior is that the process has already entered the "sleeping" state when returning from this function. Some callers use this function to generate a process and then examine the stack
trace at a certain state. However, there are steps between "tracing-stop" and "sleeping", such as "running", that may cause the stack trace from the inspected process to look slightly different when compared a stack trace from when the process once has entered the "sleeping" state. This can be seen on armhf in LP: [#2073933](https://bugs.launchpad.net/ubuntu/+source/apport/+bug/2073933).

  This commit also generally cleans up the test case and adjusts to catch the failing case from before.

### DIFF
--- a/tests/integration/test_signal_crashes.py
+++ b/tests/integration/test_signal_crashes.py
@@ -1297,7 +1297,7 @@ class T(unittest.TestCase):
             gdb_children = gdb_process.children()
             for process in gdb_children:
                 try:
-                    if process.status() == "tracing-stop":
+                    if process.status() != "sleeping":
                         continue
                     cmdline = process.cmdline()
                 except psutil.NoSuchProcess:  # pragma: no cover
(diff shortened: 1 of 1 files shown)


tests/integration/test_signal_crashes.py
```diff
@@ -1317,14 +1317,14 @@ def wait_for_gdb_sleeping_child_process(
...
-            [self.TEST_EXECUTABLE] + self.TEST_ARGS,
+        child.status.side_effect = [
+            "tracing-stop",  # child not yet started
+            "running",  # shell wrapper running
+            "sleeping",  # child ready
         ]
+        child.cmdline.side_effect = [[self.TEST_EXECUTABLE] + self.TEST_ARGS]
         with unittest.mock.patch("psutil.Process", spec=psutil.Process) as process_mock:
             process_mock.return_value.children.side_effect = [
-                [],
```
bdrung, line 1325: Why do you remove that one? This is for GDB not having started the child

tests/integration/test_signal_crashes.py
```diff
@@ -1317,14 +1317,14 @@ def wait_for_gdb_sleeping_child_process(
     def test_wait_for_gdb_sleeping_child_process(self, sleep_mock: MagicMock) -> None:
         """Test wait_for_gdb_sleeping_child_process() helper method."""
         child = MagicMock(spec=psutil.Process)
-        child.status.side_effect = ["tracing-stop", "running", "sleeping"]
-        child.cmdline.side_effect = [
-            ["/bin/sh", "-c", f"exec {self.TEST_EXECUTABLE}"],
```
bdrung, line 1320: The "shell wrapper running" has this command line. What is the reason for removing that?

tests/integration/test_signal_crashes.py
Chris-Peterson444: Ah, it wasn't clear what this was and it was having a weird affect on the state. I'll re-add it and add a comment. Thanks.

tests/integration/test_signal_crashes.py
Chris-Peterson444: Because now we only only check the process command line if the status is "sleeping" and that is the command line for the runner script or status = "running".

bdrung (approved):

PR #384, opened 2024-08-06
tests: fix flaky tests waiting for sleep command

Chris-Peterson444:
- Introduce `wait_for_sleep_cmd` test utility function to wait for a sleep command process to enter the "sleeping" state.
- Make the `create_test_process` helper functions use  `wait_for_sleep_cmd` when invoked with no arguments (the default is test executable is the sleep command) to ensure the process has entered the "sleeping" state before returning.
- Manually call this utility in `test_crash_system_slice` since it calls sleep in a more roundabout fashion via `systemd-run`. This fixes LP: [#2076186](https://bugs.launchpad.net/ubuntu/+source/apport/+bug/2076186), where armhf autopkgtests were failing due to looking for the `sleep` process before it had fully entered the "sleeping" state.
- Add a dependency on `python3-psutil` in the github CI workflow. This will also require the addition of `python3-psutil` to the Test Depends for the `system-tests` test in the packaging.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -187,8 +187,9 @@ jobs:
           ca-certificates chaos-marmosets dbus dbus-x11 dirmngr dpkg-dev gcc gdb
           gir1.2-gtk-3.0 gir1.2-wnck-3.0 git gnome-icon-theme gpg gvfs-daemons
           psmisc python3 python3-apt python3-dbus python3-gi
-          python3-launchpadlib python3-pyqt5 python3-pytest python3-pytest-cov
-          ubuntu-dbgsym-keyring ubuntu-keyring valgrind xterm xvfb
+          python3-launchpadlib python3-psutil python3-pyqt5 python3-pytest
+          python3-pytest-cov ubuntu-dbgsym-keyring ubuntu-keyring valgrind
+          xterm xvfb
       - uses: actions/checkout@v4
       - name: Start D-Bus daemon
         run: mkdir -p /run/dbus && dbus-daemon --system --fork
@@ -224,8 +225,8 @@ jobs:
           ca-certificates chaos-marmosets dbus dbus-x11 dirmngr dpkg-dev gcc gdb
           gir1.2-gtk-3.0 gir1.2-wnck-3.0 git gnome-icon-theme gpg gvfs-daemons
           pkg-config psmisc python3 python3-apt python3-dbus
-          python3-distutils-extra python3-gi python3-launchpadlib python3-pyqt5
-          python3-pytest python3-pytest-cov python3-setuptools
+          python3-distutils-extra python3-gi python3-launchpadlib python3-psutil
+          python3-pyqt5 python3-pytest python3-pytest-cov python3-setuptools
           ubuntu-dbgsym-keyring ubuntu-keyring valgrind xvfb
       - uses: actions/checkout@v4
       - name: Install
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -8,12 +8,15 @@ import os
 import pathlib
 import shutil
 import subprocess
+import time
 import unittest.mock
 import urllib.error
 import urllib.request
 from collections.abc import Callable, Generator, Iterator, Sequence
 from typing import Any
 from unittest.mock import MagicMock
+
+import psutil
 
 
 def get_init_system() -> str:
@@ -135,3 +138,22 @@ def wrap_object(
 
     with unittest.mock.patch.object(target, attribute, mocked_attribute):
         yield mock
+
+
+def wait_for_sleeping_state(pid: int, timeout: float = 5.0) -> None:
+    """Wait for sleep command to enter sleeping state."""
+    proc = psutil.Process(pid)
+    waited = 0.0
+    last_state = ""
+    while waited < timeout:
+        last_state = proc.status()
+        if last_state == "sleeping":
+            return
+
+        time.sleep(0.1)
+        waited += 0.1
+
+    raise TimeoutError(
+        f"{pid=} did not enter 'sleeping' state after {timeout=} seconds."
+        f" Got {last_state!r} instead."
+    )
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -12,7 +12,7 @@ import time
 import unittest.mock
 import urllib.error
 import urllib.request
-from collections.abc import Callable, Generator, Iterator, Sequence
+from collections.abc import Callable, Generator, Iterator, Sequence, Set
 from typing import Any
 from unittest.mock import MagicMock
 
@@ -157,3 +157,19 @@ def wait_for_sleeping_state(pid: int, timeout: float = 5.0) -> None:
         f"{pid=} did not enter 'sleeping' state after {timeout=} seconds."
         f" Got {last_state!r} instead."
     )
+
+
+def wait_for_process_to_appear(
+    process: str, already_running: Set[int], timeout: float = 5.0
+) -> int:
+    """Wait for process to appear and return its PID."""
+    waited = 0.0
+    while waited < timeout:
+        pids = pids_of(process) - already_running
+        if pids:
+            assert len(pids) == 1, f"Found more than one PID for {process!r}"
+            return pids.pop()
+        time.sleep(0.1)
+        timeout -= 0.1
+
+    raise TimeoutError(f"PID for {process!r} not found within {timeout=} seconds.")
--- a/tests/system/test_signal_crashes.py
+++ b/tests/system/test_signal_crashes.py
@@ -14,6 +14,9 @@ import sys
 import tempfile
 import time
 import unittest
+from unittest.mock import MagicMock, patch
+
+import psutil
 
 import apport.fileutils
 from tests.helper import (
@@ -304,8 +307,32 @@ class T(unittest.TestCase):
             else:
                 break
 
-        time.sleep(0.3)  # needs some more setup time
+        # sleep command needs extra time to get into "sleeping" state
+        if command == self.TEST_EXECUTABLE:
+            wait_for_sleeping_state(process.pid)
+
         return process
+
+    def test_create_test_sleep_process(self) -> None:
+        """Test create_test_process() helper method with sleep process."""
+        proc = self.create_test_process()
+        try:
+            self.assertEqual(psutil.Process(proc.pid).status(), "sleeping")
+        finally:
+            proc.kill()
+            proc.wait()
+
+    @patch("tests.helper.wait_for_sleeping_state")
+    def test_create_test_non_sleep_process(self, wait_sleep_mock: MagicMock) -> None:
+        """Test create_test_process() helper method with non-sleep process."""
+        with (
+            patch("os.access"),
+            patch("subprocess.Popen"),
+            patch("tests.system.test_signal_crashes.open"),
+        ):
+            self.create_test_process(command="echo")
+
+        wait_sleep_mock.assert_not_called()
 
     def wait_for_apport_to_finish(self, timeout_sec: float = 10.0) -> None:
         assert self.apport_path is not None
--- a/tests/system/test_signal_crashes.py
+++ b/tests/system/test_signal_crashes.py
@@ -16,7 +16,13 @@ import time
 import unittest
 
 import apport.fileutils
-from tests.helper import get_init_system, pids_of, skip_if_command_is_missing
+from tests.helper import (
+    get_init_system,
+    pids_of,
+    skip_if_command_is_missing,
+    wait_for_process_to_appear,
+    wait_for_sleeping_state,
+)
 from tests.paths import (
     get_data_directory,
     is_local_source_directory,
@@ -247,9 +253,11 @@ class T(unittest.TestCase):
             + self.TEST_ARGS,
         )
         try:
-            pids = pids_of(self.TEST_EXECUTABLE) - self.running_test_executables
-            self.assertEqual(len(pids), 1)
-            os.kill(pids.pop(), signal.SIGSEGV)
+            sleep_pid = wait_for_process_to_appear(
+                self.TEST_EXECUTABLE, self.running_test_executables
+            )
+            wait_for_sleeping_state(sleep_pid)
+            os.kill(sleep_pid, signal.SIGSEGV)
 
             self.wait_for_no_instance_running(self.TEST_EXECUTABLE)
             self.wait_for_apport_to_finish()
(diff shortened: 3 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -188,7 +188,7 @@ jobs:
           gir1.2-gtk-3.0 gir1.2-wnck-3.0 git gnome-icon-theme gpg gvfs-daemons
           psmisc python3 python3-apt python3-dbus python3-gi
           python3-launchpadlib python3-pyqt5 python3-pytest python3-pytest-cov
-          ubuntu-dbgsym-keyring ubuntu-keyring valgrind xterm xvfb
+          python3-psutil ubuntu-dbgsym-keyring ubuntu-keyring valgrind xterm xvfb
```
bdrung, line 191: Can you insert that alphabetically?

.github/workflows/ci.yaml
```diff
@@ -225,7 +225,7 @@ jobs:
           gir1.2-gtk-3.0 gir1.2-wnck-3.0 git gnome-icon-theme gpg gvfs-daemons
           pkg-config psmisc python3 python3-apt python3-dbus
           python3-distutils-extra python3-gi python3-launchpadlib python3-pyqt5
-          python3-pytest python3-pytest-cov python3-setuptools
+          python3-pytest python3-pytest-cov python3-setuptools python3-psutil
```
bdrung, line 228: Can you insert that alphabetically?

.github/workflows/ci.yaml
bdrung, line 191: suggestion: The `ci: add python3-psutil to system test dependencies` should be squashed in the commit before, because the previous commit introduces the needed change in the CI config.

tests/system/test_signal_crashes.py
```diff
@@ -246,6 +254,7 @@ def test_crash_system_slice(self) -> None:
             ]
             + self.TEST_ARGS,
         )
+        wait_for_sleep_cmd(system_run.pid)
```
bdrung, line 257: `system_run.pid` is the PID of the systemd-run process, isn't it?

tests/system/test_signal_crashes.py
```diff
@@ -246,6 +254,7 @@ def test_crash_system_slice(self) -> None:
...
         try:
             pids = pids_of(self.TEST_EXECUTABLE) - self.running_test_executables
```
bdrung, line 259: Should you while-loop for pids to detect one and then `wait_for_sleep_cmd` on that pid?

tests/system/test_signal_crashes.py
```diff
@@ -296,9 +305,31 @@ def create_test_process(self, command=None, uid=None, args=None):
...
-        time.sleep(0.3)  # needs some more setup time
+        # sleep command needs extra time to get into "sleeping" state
+        if command == self.TEST_EXECUTABLE:
+            wait_for_sleep_cmd(process.pid)
+
         return process
 
+    def test_create_test_sleep_process(self) -> None:
+        """Test create_test_process() helper method with sleep process."""
+        proc: subprocess.Popen = self.create_test_process()
```
bdrung, line 316: Explicit type hint not needed here.
```suggestion
        proc = self.create_test_process()
```

tests/system/test_signal_crashes.py
```diff
@@ -296,9 +305,31 @@ def create_test_process(self, command=None, uid=None, args=None):
...
+        finally:
+            proc.kill()
+            proc.wait()
+
+    @patch("tests.helper.wait_for_sleep_cmd")
+    def test_create_test_non_sleep_process(self, wait_sleep_mock: MagicMock) -> None:
+        """Test create_test_process() helper method with non-sleep process."""
+        with patch("os.access"):
+            with patch("subprocess.Popen"):
+                with patch("tests.system.test_signal_crashes.open"):
```
bdrung, line 328: You can put those in one `with` statement:
```suggestion
        with (
            patch("os.access"),
            patch("subprocess.Popen"),
            patch("tests.system.test_signal_crashes.open")
        ):
```
Or you can put those as patch decorators:
```python
            @patch("os.access", MagicMock())
            @patch("subprocess.Popen", MagicMock())
            @patch("tests.system.test_signal_crashes.open", MagicMock())
```

bdrung (changes requested):

tests/system/test_signal_crashes.py
Chris-Peterson444: Using patch decorators will cause the linter to complain about unused arguments, which is why I went with the context manager approach.

I had considered the single `with` statement, but I was concerned about how far back the CI would run since that's only supported in python 3.10. Although it looks like we only need to support jammy, which uses python3.10. So I'll go ahead and update it to use that.

tests/helper.py
```diff
@@ -135,3 +138,23 @@ def mocked_attribute(self: object, *args: Any, **kwargs: Any) -> Any:
 
     with unittest.mock.patch.object(target, attribute, mocked_attribute):
         yield mock
+
+
+def wait_for_sleep_cmd(pid: int, timeout: float = 5.0) -> None:
```
bdrung, line 143: suggestion: `wait_for_sleeping_state` will match more the code.

tests/system/test_signal_crashes.py
Chris-Peterson444: I think the problem here is that `pids_of` can't find the sleep command because it isn't ready (or fully started) yet.

tests/system/test_signal_crashes.py
Chris-Peterson444: Hmm you're right. I thought systemd-run would execvp the desired process but that's not quite right.

tests/helper.py
Chris-Peterson444: Agreed. Updating.

tests/system/test_signal_crashes.py
bdrung: Regarding linters: There are two solutions:
1. Specify the mock object so that the test function is not called with the mock object as parameter: `@patch("tests.system.test_signal_crashes.open", MagicMock())`
2. Check that these mocks were called.

tests/system/test_signal_crashes.py
bdrung: I would do something like that (half pseudo code):
```python
while not timeout:
   pids = pids_of(self.TEST_EXECUTABLE) - self.running_test_executables
   if pids:
      self.assertEqual(len(pids), 1)
      break
   sleep 0.1

sleep_pid = pids.pop()
wait_for_sleeping_state(sleep_pid)
os.kill(...)
...
```

tests/helper.py
```diff
@@ -135,3 +138,23 @@ def mocked_attribute(self: object, *args: Any, **kwargs: Any) -> Any:
 
     with unittest.mock.patch.object(target, attribute, mocked_attribute):
         yield mock
+
+
+def wait_for_sleeping_state(pid: int, timeout: float = 5.0) -> None:
+    """Wait for sleep command to enter sleeping state."""
+    proc = psutil.Process(pid)
+    waited = 0.0
+    last_state: str = ""
```
bdrung, line 147:
```suggestion
    last_state = ""
```

tests/unit/test_helper.py
```diff
@@ -29,3 +32,30 @@ def test_wrap_object_with_statement(self) -> None:
...
+            wait_for_sleeping_state(1234567890)
+
+        sleep_mock.assert_called_with(0.1)
+        self.assertEqual(sleep_mock.call_count, 2)
+        self.assertEqual(process_mock.return_value.status.call_count, 3)
+
+    @patch("time.sleep")
+    def test_wait_for_sleep_cmd_timeout(self, sleep_mock: MagicMock) -> None:
+        """Test wait_for_sleep_cmd() helper method times out."""
+        with patch("psutil.Process", spec=psutil.Process) as process_mock:
```
bdrung, line 55: This is inconsistent: one patch as decorator and one in a with statement. Both could be decorators.

tests/system/test_signal_crashes.py
Chris-Peterson444: Yes I like this approach. Implemented.

tests/helper.py
```diff
@@ -135,3 +138,23 @@ def mocked_attribute(self: object, *args: Any, **kwargs: Any) -> Any:
...
+
+def wait_for_sleeping_state(pid: int, timeout: float = 5.0) -> None:
+    """Wait for sleep command to enter sleeping state."""
+    proc = psutil.Process(pid)
+    waited = 0.0
+    last_state = ""
+    while waited < timeout:
+        last_state = proc.status()
+        if last_state == "sleeping":
+            break
```
bdrung, line 151: You could `return` here. Then `waited` is not needed.

Chris-Peterson444: I could remove the if statement below but we still need `waited` for the timeout loop right?

tests/helper.py
```diff
@@ -135,3 +138,23 @@ def mocked_attribute(self: object, *args: Any, **kwargs: Any) -> Any:
...
     with unittest.mock.patch.object(target, attribute, mocked_attribute):
         yield mock
+
+
+def wait_for_sleeping_state(pid: int, timeout: float = 5.0) -> None:
+    """Wait for sleep command to enter sleeping state."""
+    proc = psutil.Process(pid)
+    waited = 0.0
+    last_state = ""
+    while waited < timeout:
```
bdrung, line 148: You could carve out `wait_for_process_to_appear` from tests/integration/test_ui.py::T._gen_test_crash` and use here.

tests/helper.py
bdrung: You could count down to zero (as long as you do not need `timeout` for the error message).

tests/unit/test_helper.py
```diff
@@ -29,3 +37,83 @@ def test_wrap_object_with_statement(self) -> None:
...
+        sleep_mock.assert_called_with(0.1)
+        self.assertEqual(sleep_mock.call_count, 2)
+        self.assertEqual(process_mock.return_value.status.call_count, 3)
+
+    @patch("time.sleep")
+    @patch("psutil.Process", spec=psutil.Process)
+    def test_wait_for_sleep_cmd_timeout(
+        self, process_mock: MagicMock, sleep_mock: MagicMock
+    ) -> None:
+        """Test wait_for_sleep_cmd() helper method times out."""
```
bdrung, line 64: nitpick: The test function and docstring still refers to the old name.

tests/unit/test_helper.py
```diff
@@ -29,3 +37,83 @@ def test_wrap_object_with_statement(self) -> None:
             multiply = Multiply(7)
             self.assertEqual(multiply.multiply(6), 42)
         mock.assert_called_once_with(7)
+
+    @patch("time.sleep")
+    @patch("psutil.Process", spec=psutil.Process)
+    def test_wait_for_sleep_cmd(
+        self, process_mock: MagicMock, sleep_mock: MagicMock
+    ) -> None:
+        """Test wait_for_sleep_cmd() helper method."""
```
bdrung, line 46: nitpick: The test function and docstring still refers to the old name.

tests/unit/test_helper.py
```diff
@@ -29,3 +37,83 @@ def test_wrap_object_with_statement(self) -> None:
...
+
+    @patch("time.sleep")
+    @patch("tests.helper.subprocess.check_output")
+    def test_wait_for_process_to_appear(
+        self, check_output_mock: MagicMock, sleep_mock: MagicMock
+    ) -> None:
+        """Test wait_for_process_to_appear() helper method."""
+        check_output_mock.side_effect = [
+            "100 101 102".encode("UTF-8"),  # pre-existing processes
+            "100 101 102 103".encode("UTF-8"),  # 103 is new
```
bdrung, line 80:
```suggestion
            b"100 101 102",  # pre-existing processes
            b"100 101 102 103",  # 103 is new
```

tests/unit/test_helper.py
```diff
@@ -29,3 +37,83 @@ def test_wrap_object_with_statement(self) -> None:
...
+        ]
+        pid: int = wait_for_process_to_appear("/bin/sleep", set((100, 101, 102)))
```
bdrung, line 82:
```suggestion
        pid: int = wait_for_process_to_appear("/bin/sleep", {100, 101, 102})
```

tests/unit/test_helper.py
```diff
@@ -29,3 +37,83 @@ def test_wrap_object_with_statement(self) -> None:
...
+        sleep_mock.assert_called_with(0.1)
+        self.assertEqual(sleep_mock.call_count, 1)
+
+    @patch("time.sleep")
+    @patch("tests.helper.subprocess.check_output")
+    def test_wait_for_process_to_appear_timeout(
+        self, check_output_mock: MagicMock, sleep_mock: MagicMock
+    ) -> None:
+        """Test wait_for_process_to_appear() helper method times out."""
+        check_output_mock.return_value = "".encode("UTF-8")
```
bdrung, line 94:
```suggestion
        check_output_mock.return_value = b""
```

tests/unit/test_helper.py
```diff
@@ -29,3 +37,83 @@ def test_wrap_object_with_statement(self) -> None:
...
+    def test_wait_for_process_to_appear_multiple(
+        self, check_output_mock: MagicMock, sleep_mock: MagicMock
+    ) -> None:
+        """Test wait_for_process_to_appear() helper method raises AssertionError.
+
+        wait_for_process_to_appear expects to find only one PID.
+        """
+        check_output_mock.side_effect = [
+            "100 101 102".encode("UTF-8"),  # pre-existing processes
+            "100 101 102 103 104".encode("UTF-8"),  # 103 and 104 are new
```
bdrung, line 112:
```suggestion
            b"100 101 102",  # pre-existing processes
            b"100 101 102 103 104",  # 103 and 104 are new
```

tests/unit/test_helper.py
```diff
@@ -29,3 +37,83 @@ def test_wrap_object_with_statement(self) -> None:
...
+        ]
+        check_output_mock.return_value = "".encode("UTF-8")
```
bdrung, line 114: return_value and side_effect?

tests/unit/test_helper.py
```diff
@@ -29,3 +37,83 @@ def test_wrap_object_with_statement(self) -> None:
...
+        with self.assertRaises(AssertionError):
+            wait_for_process_to_appear("/bin/sleep", set((100, 101, 102)))
```
bdrung, line 116:
```suggestion
            wait_for_process_to_appear("/bin/sleep", {100, 101, 102})
```

tests/unit/test_helper.py
Chris-Peterson444: Thanks, this was an oversight. Technically side_effect takes precedence so the return_value line is a no-op but I will remove it.

bdrung (approved):

PR #385, opened 2024-08-08
test: Add more type hints

bdrung:
Add more type hints to tests.

### DIFF
--- a/tests/integration/test_report.py
+++ b/tests/integration/test_report.py
@@ -17,6 +17,7 @@ import textwrap
 import time
 import unittest
 import unittest.mock
+from typing import IO
 from unittest.mock import MagicMock
 
 import apport.packaging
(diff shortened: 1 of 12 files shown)


tests/integration/test_report.py
```diff
@@ -841,7 +842,7 @@ def test_add_gdb_info_exe_missing(self) -> None:
     @unittest.mock.patch(
         "apport.hookutils._root_command_prefix", MagicMock(return_value=[])
     )
-    def test_add_zz_parse_segv_details(self) -> None:
```
schopin-pro, line 844: question: why do we *drop* type hints here?

Not complaining, just wondering.

schopin-pro (approved):
LGTM overall but there's a lack of consistency over the `-> None`

tests/integration/test_report.py
bdrung: Thanks for noticing. That was unintentional and unwanted. Restored those type hints.

PR #387, opened 2024-08-08
apt_dpkg: Map mirror to correct URL for the architecture

bdrung:
The Ubuntu archive is split. The primary mirrors only has packages for the amd64 and i386 architectures. The ports mirrors host all remaining architectures.

So introduce a mapping function that correctly changes the URL in case known problematic ones were found.

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -125,6 +125,25 @@ def _load_all_sources(apt_dir: str) -> list[apt_sl.Deb822SourceEntry | SourceEnt
     return sources
 
 
+def _map_mirror_to_arch(uri: str, target_arch: str) -> str:
+    """Map the givin archive mirror URI to a valid one for the givin architecture.
+
+    The Ubuntu archive is split. The primary mirrors only has packages
+    for the amd64 and i386 architectures. The ports mirrors host all
+    remaining architectures.
+    """
+    if target_arch in {"amd64", "i386"}:
+        ports_match = re.match("http://([a-z.]+)?ports.ubuntu.com/ubuntu-ports/?", uri)
+        if ports_match:
+            return f"http://{ports_match.group(1)}archive.ubuntu.com/ubuntu"
+        return uri
+
+    primary_match = re.match("http://([a-z.]+)?archive.ubuntu.com/ubuntu/?$", uri)
+    if primary_match:
+        return f"http://{primary_match.group(1)}ports.ubuntu.com/ubuntu-ports"
+    return uri
+
+
 def _read_mirror_file(uri: str) -> list[str]:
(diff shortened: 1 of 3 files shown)


apport/packaging_impl/apt_dpkg.py
```diff
@@ -125,6 +125,25 @@ def _load_all_sources(apt_dir: str) -> list[apt_sl.Deb822SourceEntry | SourceEnt
...
 
 
+def _map_mirror_to_arch(uri: str, arch: str) -> str:
+    """Map the givin archive mirror URI to a valid one for the givin architecture.
+
+    The Ubuntu archive is split. The primary mirrors only has packages
+    for the amd64 and i386 architectures. The ports mirrors host all
+    remaining architectures.
+    """
+    if arch in {"amd64", "i386"}:
```
schopin-pro, line 135: thought: it's actually probably better to use a tuple here in terms of raw perf, as a set is more expensive to build and the lookup might not be faster than simple iteration on such a small size.

schopin-pro, line 135: question: shouldn't this be the opposite? We want ports for non-x86 architectures!

schopin-pro (commented):
I'm thoroughly confused by this.

apport/packaging_impl/apt_dpkg.py
bdrung: Good point. I'll change that.

schopin-pro: LOL, the linter disagrees with me here.

bdrung: Pylint complained about it. So I decided to check the performance claims:
```
In [19]: arch="amd64"

In [20]: timeit.timeit('arch in ("amd64", "i386")', number=100000000, globals=globals())
Out[20]: 1.1992814919940429

In [21]: timeit.timeit('arch in {"amd64", "i386"}', number=100000000, globals=globals())
Out[21]: 1.39064237200364

In [22]: arch="armhf"

In [23]: timeit.timeit('arch in ("amd64", "i386")', number=100000000, globals=globals())
Out[23]: 2.203924414003268

In [24]: timeit.timeit('arch in {"amd64", "i386"}', number=100000000, globals=globals())
Out[24]: 1.2740627159946598
```
The tuple is not necessarily faster. So I would stick with the set.

bdrung: The logic is correct: we search for the ports URL to replace it with the non-ports URL for x64 architectures.

schopin-pro: Alright, then may I suggest renaming "arch" to "target_arch"?

bdrung: Yes, that makes it clearer. Changed.

PR #399, opened 2024-10-01
Fix hang when cancelling/closing Apport

bdrung:
Canceling the upload of problem information after sending information acquired with `ubuntu-bug [package name]` causes the upload dialog to become unresponsive.

The main thread is terminated when cancelling the crash collection/uploading, but the spawned background thread will continue running.

Mark the spawned background threads as daemon threads. This will stop those threads when the main tread terminates.

Bug-Ubuntu: https://launchpad.net/bugs/1537310

### DIFF
--- a/apport/REThread.py
+++ b/apport/REThread.py
@@ -17,14 +17,18 @@ import threading
 
 
 class REThread(threading.Thread):
-    """Thread with return values and exception propagation."""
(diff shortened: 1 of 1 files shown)


schopin-pro:
LGTM. The coverage issue seems unrelated, probably some flakiness depending on the exact state of the archive.

PR #401, opened 2024-11-08
test: do not check for exact encoded gzip data in test_write_file

bdrung:
The test case `test_write_file` fails on s390x from time to time due to the encoded chunks have a different on-disk representation. Instead of the expected one compressed chunk

```
File: base64
 H4sICAAAAAAC/0ZpbGUA
 c3RyxIAMcBAFAK/2p9MfAAAA
```

there could be two compressed chunks:

```
File: base64
 H4sICAAAAAAC/0ZpbGUA
 cnTChAxwEA==
 BRgAr/an0x8AAAA=
```

The raw data is identical, but the zlib compressed data on disk differs! So make the test case less picky by just checking that the data is compressed and that the decompressed data is correct.

Bug: https://launchpad.net/bugs/2076269

### DIFF
--- a/tests/integration/test_problem_report.py
+++ b/tests/integration/test_problem_report.py
@@ -228,18 +228,12 @@ class T(unittest.TestCase):
             out = io.BytesIO()
             pr.write(out)
 
-            self.assertEqual(
-                out.getvalue().decode(),
-                textwrap.dedent(
-                    """\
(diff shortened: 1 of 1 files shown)


schopin-pro:
To be clear, this is a workaround rather than a fix, but from the narrow PoV of apport development I'm fine with that.

PR #404, opened 2024-11-22
ci: Drop testing Ubuntu 22.04 "jammy" and increase required Python version to at least 3.11

bdrung:
Ubuntu 24.04 "noble" is the latest Ubuntu LTS version.

Debian 12 "Bookworm" ships Python 3.11 and Ubuntu 24.04 "noble" ships Python 3.12. So bump the baseline after dropping Ubuntu 22.04 "jammy".


bdrung:
This PR should be ready. The reduced codecov/project can be ignored IMO.

PR #407, opened 2024-11-22
chore: drop obsolete pylint workarounds

bdrung:
Drop obsolete pylint workarounds that pylint 3.2.7 on Ubuntu 24.10 "oracular" does not complain about any more.

### DIFF
--- a/data/apport
+++ b/data/apport
@@ -15,9 +15,6 @@ See https://wiki.ubuntu.com/Apport for details."""
 
 # pylint: disable=too-many-lines
 
-# pylint fails to import the apport module, because it has the same name.
-# See bug https://github.com/PyCQA/pylint/issues/7093
-# pylint: disable=c-extension-no-member,no-name-in-module,not-callable
 # TODO: Address following pylint complaints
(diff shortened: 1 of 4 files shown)


tests/integration/test_ui.py
```diff
@@ -676,8 +676,6 @@ def test_file_report_http_error(self) -> None:
         self.ui = UserInterfaceMock()
         self.ui.report = self.report
         with unittest.mock.patch.object(self.ui.crashdb, "upload") as upload_mock:
-            # False positive for hdrs in urllib.error.HTTPError
-            # See https://github.com/python/typeshed/issues/10092
```
schopin-pro, line 680: question: you're removing a standard comment, no modification of either code or pylint hints?

schopin-pro (approved):
LGTM, just a simple question on one of the changes.

tests/integration/test_ui.py
bdrung: You are right. That removal was part of testing what can be dropped. I restored that comment.

PR #411, opened 2024-12-04
Snap package hooks

matthew-hagemann:
Adds support for `apport` to search `/snap/<snap-name>/current/apport/package-hooks` for package hooks.

Currently there are several package-hooks the desktop team have [in `apport` itself](https://git.launchpad.net/~ubuntu-core-dev/ubuntu/+source/apport/tree/debian/package-hooks/ubuntu-desktop-bootstrap.py).  

This will allow for easier apport integration for snap packagers, and allow us to move special case
snap hooks out of `/usr/share/apport/package-hooks/` in the future, making them easier to iterate on by including package hooks in the snap packaging.

### DIFF
--- a/tests/integration/test_report.py
+++ b/tests/integration/test_report.py
@@ -1559,6 +1559,47 @@ int main() { return f(42); }
             apport.report.PACKAGE_HOOK_DIR = orig_package_hook_dir
             apport.report._opt_dir = orig_opt_dir
 
+    def test_add_hooks_info_snap(self) -> None:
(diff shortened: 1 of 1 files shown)


bdrung:
Thanks for your contribution, but there is a security issue: The snap hook is run outside of its container, which would defeat the purpose of snaps.

bdrung (changes requested):
We cannot trust the hook provided by the snap. If we want to run hooks from snaps, those hooks need to run in the context of the snap.

PR #413, opened 2024-12-06
hookutils.attach_mac_events pkexec refactoring

schopin-pro:
This should reduce the number of `pkexec` calls to 1 for this function, no matter the branching.

### DIFF
--- a/tests/unit/test_hookutils.py
+++ b/tests/unit/test_hookutils.py
@@ -374,3 +374,10 @@ class TestHookutils(unittest.TestCase):
             )
 
             self.assertEqual(match_content, "test\ncontent\n")
+
+    @staticmethod
+    @unittest.mock.patch("subprocess.run")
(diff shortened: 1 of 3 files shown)


schopin-pro:
I think we can disregard codecov, the changed code will be covered by #412

tests/unit/test_hookutils.py
```diff
@@ -374,3 +374,10 @@ def test_recent_syslog_long_process(self) -> None:
             )
 
             self.assertEqual(match_content, "test\ncontent\n")
+
+    @staticmethod
+    @unittest.mock.patch("subprocess.run")
+    def test_execute_multiple_root_commands_no_commands(run_mock):
```
bdrung, line 380:
```suggestion
    def test_execute_multiple_root_commands_no_commands(run_mock: MagicMock) -> None:
```

bdrung:
> I think we can disregard codecov, the changed code will be covered by #412

I agree.

PR #415, opened 2024-12-06
hookutils: new helper execute_multiple_root_commands

schopin-pro:
Very much like attach_root_command_outputs except that it simply stores the outputs in a dict rather than modifying an existing report. Very useful for post-processing.

### DIFF
--- a/apport/hookutils.py
+++ b/apport/hookutils.py
@@ -28,6 +28,7 @@ import subprocess
 import sys
 import tempfile
 import warnings
+from collections.abc import Mapping
 
 import apport.fileutils
 from apport.packaging_impl import impl as packaging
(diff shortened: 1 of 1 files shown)


apport/hookutils.py
```diff
@@ -520,6 +523,10 @@ def attach_root_command_outputs(report, command_map):
     This is preferable to using root_command_output() multiple times, as that
     will ask for the password every time.
     """
+    if not command_map:
+        return {}
```
bdrung, line 527: Is there a useful case for that?

schopin-pro: Yes. The following pattern is fairly trivial to end up with:

```python
missing_data = {}
if not data['foo']:
    missing_data['foo'] = 'bar'
if not data['baz']:
    missing_data['baz'] = 'foo'
data.update(execute_multiple_root_commands(missing_data))
```
That way you don't need to worry about it as a caller, and having an early return means the user won't be prompted to run an empty script.

bdrung: Okay, valid point. So can we have a unit test for that?

bdrung: As mentioned on chat: The following PRs will cover this code path. So no separate test case needed.

bdrung (approved):

PR #417, opened 2024-12-09
test: convert test_attach_mac_events into a unit test

bdrung:
Convert the `test_attach_mac_events` integration test into a unit test. The unit test always sets `AuditLog` and `KernLog` to avoid calling commands. Mock `execute_multiple_root_commands` to ensure that no command is called.


tests/unit/test_hookutils.py
```diff
@@ -190,6 +191,160 @@ def test_attach_journal_errors_without_date(self, run_mock: MagicMock) -> None:
             ["journalctl", "--priority=warning", "-b", "--lines=1000"],
         )
 
+    @unittest.mock.patch("subprocess.run")
```
schopin-pro, line 194: issue: subprocess.run is the wrong thing to mock out, you're breaking encapsulation.

If you write a unit test for a function, you shouldn't rely on implementation details of *other* functions to test out its behavior.

If I switch out the implementation of `execute_multiple_root_commands` to use `subprocess.Popen()`, the only unit tests that should be need to change are that function's tests.

Yes, that means the tests you're writing/amending won't be as easy to write, and would have needed to change before/after #414, but that's fine since we're actually *changing the tested function*.

schopin-pro (changes requested):
NACK on the current approach, it's breaking encapsulation for no real gain that I can see.

tests/unit/test_hookutils.py
bdrung: Changed to mock `execute_multiple_root_commands` instead.

PR #422, opened 2024-12-11
Remove apt key

julian-klode:
apt-key has been removed in apt 2.9.17. Fetch the signing key directly from launchpad and store it in a .asc file.

Change the test cases to compare the ASCII-armored keys rather than check gpg output.


apport/packaging_impl/apt_dpkg.py
```diff
@@ -1909,33 +1909,15 @@ def _build_apt_sandbox(
...
-                ]
-
-                if subprocess.call(argv) != 0:
-                    apport.logging.warning(
-                        "Unable to import key for %s", ppa_archive_url
-                    )
+                with open(
+                    os.path.join(trusted_d, f"{origin}.asc"), "w", encoding="utf-8"
+                ) as key_out:
+                    key_out.write(ppa_key)
```
bdrung, line 1920: You could use pathlib to do the same with a little less code. Should look like this (more or less):
```suggestion
                key_file = pathlib.Path(trusted_d) / f"{origin}.asc"
                key_file.write_text(ppa_key, encoding="utf-8")
```

apport/packaging_impl/apt_dpkg.py
```diff
@@ -1909,33 +1909,13 @@ def _build_apt_sandbox(
...
-                    continue
-                try:
-                    signing_key_fingerprint = ppa_info["signing_key_fingerprint"]
-                except IndexError:
-                    apport.logging.warning(
-                        "Error: can't find signing_key_fingerprint at %s",
-                        ppa_archive_url,
-                    )
+                ppa_key = self.json_request(
+                    ppa_archive_url + "?ws.op=getSigningKeyData"
```
bdrung, line 1913:
```suggestion
                    f"{ppa_archive_url}?ws.op=getSigningKeyData"
```

tests/system/test_packaging_apt_dpkg.py
```diff
@@ -752,33 +753,20 @@ def test_create_sources_for_a_named_ppa(
...
-            ),
-        ],
-        check=True,
-        stdout=subprocess.PIPE,
+    actual_file = (
+        pathlib.Path(rootdir)
+        / "etc"
+        / "apt"
+        / "trusted.gpg.d"
+        / "LP-PPA-daisy-pluckers-daisy-seeds.asc"
```
bdrung, line 761: Maybe splitting that into two lines could make it more readable:
```python
trusted_gpg_d = pathlib.Path(rootdir) / "etc" / "apt" / "trusted.gpg.d"
actual_file = trusted_gpg_d / "LP-PPA-daisy-pluckers-daisy-seeds.asc"
```

bdrung (commented):
Thanks. The change looks good. Some suggestions:
* Maybe add a fixes line pointing to the commit that broke
* Maybe invert the order of both commit (in case the first one will break without the second one)
* It looks like no test case uses `gpg` directly any more. So gpg can be dropped from `.github/workflows/ci.yaml`.

bdrung (approved):

PR #425, opened 2024-12-21
problem_report: introduce ProblemReportValue type alias

bdrung:
* problem_report: reuse `self[key]` in `sorted_items` (for the following commits)
* apport-cli: only calculate length for strings to show
* problem_report: sorted_items can also return `CompressedFile`
* Add a type alias for the value type of the `ProblemReport` class.

### DIFF
--- a/bin/apport-cli
+++ b/bin/apport-cli
@@ -159,12 +159,12 @@ class CLIUserInterface(apport.ui.UserInterface):
         max_show = 1000000
         for key, value in self.report.sorted_items():
             details += f"== {key} =================================\n"
-            # string value
-            keylen = len(value)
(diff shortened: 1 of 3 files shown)


schopin-pro (changes requested):
NACK, implementing `len()` for CompressedFile is the wrong solution here. As proposed, it would give use the compressed length where CompressedValue gives us the decompressed length, meaning there's a semantic discrepancy. If you look at where said length would be used, you can see that it can be worked around there to only compute the length on strings and fold the rest as binary data.

bdrung:
Agreed. I dropped the `CompressedFile.__len__` commit and changed apport-cli instead.

PR #430, opened 2025-01-17
Expect pid parameter of Report.add_proc_info to be of type int

bdrung:
* test: use `tempfile.TemporaryDirectory` in `test_add_proc_info`
* unkillable_shutdown: pass `pid` around as `int`
* Add type hints to all callers of `Report.add_proc_info`
* report: expect `pid` parameter of `add_proc_info` to be of type int

### DIFF
--- a/tests/integration/test_report.py
+++ b/tests/integration/test_report.py
@@ -8,6 +8,7 @@ import atexit
 import grp
 import io
 import os
+import pathlib
 import re
 import shutil
 import subprocess
@@ -137,7 +138,7 @@ class T(unittest.TestCase):
         self.assertNotIn("UserGroups", report)
         geteuid_mock.assert_called_once_with()
 
-    def test_add_proc_info(self):
+    def test_add_proc_info(self) -> None:
         # TODO: Split into separate test cases
         # pylint: disable=too-many-statements
         """add_proc_info()."""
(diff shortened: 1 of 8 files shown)


tests/integration/test_report.py
```diff
@@ -236,31 +237,30 @@ def test_add_proc_info(self):
...
-            pr = apport.report.Report()
-            pr.add_proc_info(pid=process.pid)
-            process.communicate(b"\n")
-        self.assertEqual(pr["ExecutablePath"], testscript)
-        self.assertEqual(
-            int(pr["ExecutableTimestamp"]), int(os.stat(testscript).st_mtime)
-        )
-        os.unlink(testscript)
+        with tempfile.TemporaryDirectory(prefix="apport_") as tempdir:
+            testscript = f"{tempdir}/testscript.py"
```
schopin-pro, line 241: suggestion:
```python
testscript = pathlib.Path(tempdir) / "testscript.py"
```

(would need to change some stuff down the line)

schopin-pro (approved):
Just one suggestion.

tests/integration/test_report.py
bdrung: Good idea. I changed testscript to fully use pathlib.

PR #431, opened 2025-01-17
test: use unittest.mock.patch for mocking methods

bdrung:
Use `unittest.mock.patch.object` for mocking methods in the test cases to make mypy happy.

### DIFF
--- a/tests/integration/test_ui.py
+++ b/tests/integration/test_ui.py
@@ -1614,7 +1614,7 @@ class T(unittest.TestCase):
             pwd.getpwuid = orig_getpwuid
             os.getuid = orig_getuid
 
-    def test_run_crash_known(self):
+    def test_run_crash_known(self) -> None:
         """run_crash() for already known problem"""
         r = self._gen_test_crash()
         report_file = os.path.join(apport.fileutils.report_dir, "test.crash")
@@ -1624,8 +1624,9 @@ class T(unittest.TestCase):
         # known without URL
         with open(report_file, "wb") as f:
             r.write(f)
-        self.ui.crashdb.known = lambda r: True
-        self.ui.run_crash(report_file)
+        with patch.object(self.ui.crashdb, "known", return_value=True) as mock:
+            self.ui.run_crash(report_file)
+        mock.assert_called()
         assert self.ui.report
         self.assertEqual(self.ui.report["_KnownReport"], "1")
         self.assertEqual(self.ui.msg_severity, "info")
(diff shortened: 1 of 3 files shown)


schopin-pro (approved):
LGTM, just one suggestion to improve it s'more.

tests/integration/test_ui.py
```diff
@@ -2430,52 +2430,45 @@ def run(report, ui):
...
         """can_examine_locally() for a crash report"""
         self.ui.load_report(self.report_file.name)
 
         orig_path = os.environ["PATH"]
-        orig_fn = self.ui.ui_has_terminal
         try:
-            self.ui.ui_has_terminal = lambda command: True
             os.environ["PATH"] = ""
-            self.assertEqual(self.ui.can_examine_locally(), False)
+            with patch.object(self.ui, "ui_has_terminal", return_value=True):
```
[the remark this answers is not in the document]
bdrung: I added `mock.assert_called` for all tests, except for mocking ui_has_terminal, because these tests do not call the mock. I will investigate that and fix in followup PR.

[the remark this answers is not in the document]
bdrung: Added assertions for mocking `ui_has_terminal` where applicable. Some of the cases do not care if the code reaches the `ui_has_terminal` call.

tests/integration/test_ui.py
```diff
@@ -2430,52 +2430,47 @@ def run(report, ui):
...
         del self.ui.report["CoreDump"]
 
-        orig_fn = self.ui.ui_has_terminal
-        try:
-            self.ui.ui_has_terminal = lambda command: True
+        with patch.object(self.ui, "ui_has_terminal", return_value=True) as mock:
             self.assertEqual(self.ui.can_examine_locally(), False)
-        finally:
-            self.ui.ui_has_terminal = orig_fn
+        mock.assert_not_called()
```
schopin-pro, line 2478: question: why are we mocking in the first place then?

schopin-pro (commented):
issue: Why constrain the assertions to `assert_called_once()` rather than simple `assert_called` if there are no side-effects associated with the mocked functions? It makes the tests more fragile.

tests/integration/test_ui.py
bdrung: To ensure that `can_examine_locally` does not return `False` due to `ui_has_terminal` returning `False`.

bdrung:
Valid point. Replaced `assert_called_once` with `simple assert_called`.

tests/integration/test_ui.py
schopin-pro: By that reasoning, there are thousands of other program properties that could potentially change `can_examine_locally`, shall we mock them all?

bdrung: Okay. Removed the mocking from `test_can_examine_locally_nocrash`.

PR #432, opened 2025-01-17
test: support relative path in do_crash

bdrung:
The test `test_unpackaged_script` is skipped, because `do_crash` does not support relative executable path any more:

```
SKIPPED [1] tests/integration/test_signal_crashes.py:382: ./myscript is not executable
```

Determine absolute path for `os.access` and `read_shebang` in `do_crash` when `cwd` is specified.

Also add a commit for using pathlib.

Fixes: 7ae2c0b541a9 ("tests: Convert signal crashes tests into integration tests")

### DIFF
--- a/tests/integration/test_signal_crashes.py
+++ b/tests/integration/test_signal_crashes.py
@@ -1118,6 +1118,7 @@ class T(unittest.TestCase):
         hook_before_apport: Callable | None = None,
         expect_report: bool = True,
         via_socket: bool = False,
+        cwd: str | None = None,
         **kwargs: typing.Any,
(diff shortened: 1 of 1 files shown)


schopin-pro (approved):
LGTM.

nitpick: Your Fixes field is wrong and/or incomplete, the root cause of the bug is https://github.com/canonical/apport/commit/7ae2c0b541a91bf42fcb416c874664810477dc3c, the one you're mentioning is only the first time we're using that particular buggy codepath.

bdrung:
Valid point. I changed the "Fixes" line.

PR #435, opened 2025-02-04
Clear APT::Architectures when setting architecture

julian-klode:
When we set the architecture for our faked chroot, we need to unset foreign architectures as well, otherwise wew end up fetching e.g. i386 packages from ports.


julian-klode:
A fix for this on the APT side is in https://salsa.debian.org/apt-team/apt/-/merge_requests/445 but this is probably still the right thing to merge here given it's a library function.

bdrung (approved):

PR #440, opened 2025-02-18
report: introduce helper _provide_uncompressed_coredump_file()

bdrung:


apport/report.py
```diff
@@ -1858,6 +1858,24 @@ def anonymize(self) -> None:
                     else:
                         self[k] = pattern.sub(repl, self[k])
 
+    def _provide_uncompressed_coredump_file(self) -> str:
+        coredump = self["CoreDump"]
+        if hasattr(coredump, "find"):
```
schopin-pro, line 1863: issue(doc): please document what that means in concrete terms

"find" is a fairly generic method name, I have no idea what types you're expecting here!

schopin-pro (changes requested):

apport/report.py
bdrung: I have a pending branch that replaces that `hasattr` by `isinstance`, but please allow that change to land in a future merge request.

schopin-pro: Well, meanwhile I'm left with a bit of code I don't understand, which is not exactly the ideal position to merge stuff.

bdrung: `hasattr(coredump, "find")` is equivalent to `isinstance(coredump, bytes | str)`. `str` is not probably not support (I will test that assumption).

bdrung: I pushed the two additional commits to make the code more readable.

schopin-pro: Much appreciated, it makes a lot more sense now :)

PR #442, opened 2025-02-19
test: add more type hints to integration tests

bdrung:


schopin-pro (approved):
The PR does a bit more than what it says on the tin, but I'll take it :)

PR #443, opened 2025-02-19
ui: drop unneeded int conversion in run_hang

bdrung:
The method `UserInterface.run_hang` requires `pid` to be of type `int`.

### DIFF
--- a/apport/ui.py
+++ b/apport/ui.py
@@ -915,8 +915,7 @@ class UserInterface:
             self.run_symptom()
             return True
         if self.args.hanging:
-            self.run_hang(self.args.pid)
-            return True
+            return self.run_hang(self.args.pid)
         if self.args.filebug:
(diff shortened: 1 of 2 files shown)


apport/ui.py
```diff
@@ -545,7 +545,7 @@ def finish_hang(f: str) -> None:
         apport.fileutils.mark_report_upload(f)
         apport.fileutils.mark_report_seen(f)
 
-    def run_hang(self, pid: int) -> bool:
+    def run_hang(self, pid: int) -> None:
```
schopin-pro, line 548: question: isn't that a public API?

bdrung: I guess that it is not called by someone outside, but I don't know. I looked it up: the method isn't a recent invention. So we can stay on the safe side and keep the `bool` return.

PR #454, opened 2025-02-21
run-linters: check Python code in debian/ if present

bdrung:
The Ubuntu package ships package hooks in `debian/package-hooks`. Check Python code in the `debian` directory if this directory exists.


schopin-pro:
That's a NACK from me. Running linters on package builds just makes it harder to cherry-pick patches.

bdrung:
The Debian package runs `tests/run-linters --errors-only` during package build and nothing during autopkgtest.

I want this change to easily run the tests on my machine.

schopin-pro:
You want it, you merge it :)

PR #459, opened 2025-02-27
Apply uncontroversial ruff format changes

bdrung:
Black and `ruff format` disagree how to format some multi-line Python code, but ruff is more aggressive in unwrapping lines. So apply the uncontroversial ruff format changes:

```
ruff format
black -C .
```

### DIFF
--- a/apport/crashdb_impl/debian.py
+++ b/apport/crashdb_impl/debian.py
@@ -8,7 +8,6 @@
 # option) any later version.  See http://www.gnu.org/copyleft/gpl.html for
 # the full text of the license.
 
-
 import email.mime.text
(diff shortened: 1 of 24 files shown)


bdrung:
IMO codecov/patch can be ignored here since it failed to detect all modified lines.

schopin-pro (changes requested):
NACK, we pick one set of rules enforced by ONE TOOL and we stick with it. If you want to add  to those rules in your own commits go for it, but don't enforce it globally unless you have (full) automation to back it up.

bdrung:
Then let's wait for `ruff format` and `black` to not disagree.

PR #464, opened 2025-03-12
ui: update_report: fix source/binary confusion (LP: #2102147)

schopin-pro:
Rather than relying on binary packages being named the same as their
source package, we actually split the two of them. There's an underlying
assumption that we're reporting to a backend that cares about source
packages, so that's what we're basing it all on, by adding a field to
list the binary packages actually installed from that source package.

It might not work perfectly (the Package field will be marked with "not
installed" for glibc, for instance), but it's still much better than the
current state.

Bug-Ubuntu: https://bugs.launchpad.net/ubuntu/+source/apport/+bug/2102147

### DIFF
--- a/apport/ui.py
+++ b/apport/ui.py
@@ -777,27 +777,38 @@ class UserInterface:
 
         # list of affected source packages
         self.report = apport.Report("Bug")
+        # FIXME: for now we assume this is a binary package because it's just easier
+        # and it didn't work with a source package anyway.
         if self.args.package:
-            pkgs = [self.args.package.strip()]
+            binary = self.args.package.strip()
+            pkgs = {packaging.get_source(binary): set((binary,))}
         else:
-            pkgs = self.crashdb.get_affected_packages(self.args.update_report)
+            pkgs = {
+                src: packaging.get_installed_binaries(src)
+                for src in self.crashdb.get_affected_packages(self.args.update_report)
+            }
 
         info_collected = False
-        for p in pkgs:
+        assert self.report is not None
+        for src, binaries in pkgs.items():
             # print(f"Collecting apport information for source package {p}...")
-            self.cur_package = p
-            self.report["SourcePackage"] = p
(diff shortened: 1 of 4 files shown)


schopin-pro:
Ugh, apparently the tests assume that -p is for the source package. This entire thing is just sooooo confusing.

apport/ui.py
```diff
@@ -777,27 +777,38 @@ def run_update_report(self) -> bool:
...
+        for src, binaries in pkgs.items():
             # print(f"Collecting apport information for source package {p}...")
-            self.cur_package = p
-            self.report["SourcePackage"] = p
-            self.report["Package"] = p  # no way to find this out
+            assert binaries
+            self.cur_package = src
+            self.report["SourcePackage"] = src
+            self.report["Package"] = src  # Slight abuse of the field but it's mandatory
+            self.report["InstalledBinaries"] = " ".join(sorted(binaries))
```
bdrung, line 799: We probably also want to collect the version of those packages (similar to what `Dependencies` does).

apport/ui.py
```diff
@@ -777,27 +777,38 @@ def run_update_report(self) -> bool:
...
 
             # we either must have the package installed or a source package
             # hook available to collect sensible information
             try:
-                packaging.get_version(p)
+                packaging.get_version(next(iter(binaries)))
```
bdrung, line 804: That will only return one of the binary packages and is not stable.

apport/ui.py
```diff
@@ -777,27 +777,38 @@ def run_update_report(self) -> bool:
...
             except ValueError:
```
bdrung, line 805: question: Can we still reach that exception?

apport/ui.py
schopin-pro: I was going to answer that we don't need it since it's already known, but actually that's a really good point, that would show mixed-version weirdness.

PR #465, opened 2025-03-12
crashdb: Import sqlite3 at the top level

bdrung:
The Python module `sqlite3` is part of the Python standard library.

### DIFF
--- a/apport/crashdb.py
+++ b/apport/crashdb.py
@@ -15,6 +15,7 @@
 import functools
 import os
 import shutil
+import sqlite3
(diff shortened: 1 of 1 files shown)


schopin-pro (changes requested):
NACK, the lint is actually *wrong* here. Only importing stuff when you need them is a *good* thing, it reduces load time for all those other use cases.

bdrung:
Valid point. You get https://github.com/canonical/apport/pull/471 instead.

PR #477, opened 2025-03-14
Rename apport/packaging.py to apport/package_info.py

bdrung:
`import apport.packaging` will not import `apport/packaging.py` but `packaging` from `apport/__init__.py` instead. To avoid this name clash name the Python file identical to the class it defines.

It is expected that no Python code outside of Apport uses `from apport.packaging import PackageInfo`.

The file `apport/packaging.py` is kept as empty file with a long description why it is still needed. Avoid looking at this file. You have been warned!

Fixes: 8763b83047a9 ("Changed the packaging backend import, [...]")


bdrung:
The failing codecov/patch can be ignored.

schopin-pro:
Please resolve the conflicts.

bdrung:
Rebased and conflict resolved.

PR #481, opened 2025-03-21
Deprecate logging functions in base apport module

bdrung:
Deprecate using the logging functions directly from the base `apport` module. Users should import the logging functions from `apport.logging` instead.


schopin-pro:
Very rough benchmarking on the lazyloading (basically looping ofter `python3 -X importtime -c 'import apport' a hundred times and averaging the result) gives me a "before" of 71267.02970297029 and an "after" of 67990.25742574257, so I'm sold.

PR #483, opened 2025-03-21
Address most pylint's invalid-name complains

bdrung:
Address most pylint's invalid-name complains. See individual commits for details.

### DIFF
--- a/apport/crashdb_impl/launchpad.py
+++ b/apport/crashdb_impl/launchpad.py
@@ -12,7 +12,7 @@
 
 # pylint: disable=too-many-lines
 # TODO: Address following pylint complaints
-# pylint: disable=invalid-name,missing-function-docstring
+# pylint: disable=missing-function-docstring
 
 import atexit
 import email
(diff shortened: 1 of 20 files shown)


apport/crashdb_impl/launchpad.py
```diff
@@ -1158,7 +1158,7 @@ def _filter_tag_names(tags: str) -> str:
 # Launchpad storeblob API (should go into launchpadlib, see LP #315358)
 #
 
-_https_upload_callback = None
+_HTTPS_UPLOAD_CALLBACK = None
```
schopin-pro, line 1161: issue: this is NOT a constant.

It's an ugly hack, not a constant. Please don't try to pass it off as one.

schopin-pro (changes requested):
*Almost* there.

apport/crashdb_impl/launchpad.py
bdrung: I dropped that part of the patch (and found DEFAULT_CREDENTIALS_PATH which is a constant).

PR #485, opened 2025-03-25
report: add support for YAQL expressions in bug patterns

ogayot:
Bug patterns have historically only supported executing regular expressions. This is ok for log files and simple strings, but when we have access to more structured data (i.e., JSON or YAML) in bug reports, having the ability to find elements in such files is very useful.

In particular, this would be beneficial for Subiquity, where we have a large number of bug reports that we currently de-duplicate manually. We want to rely more on automation for such tasks.

In addition to regular expressions, we now support YAQL expressions [1] in bug patterns. They can be specified using the following syntax:

```xml
<pattern url="https://launchpad.net/bugs/XXX">
  <re key="package">subiquity</re>
  <yaql key="ProbeData">
      [ YAQL expression ]
  </yaql>
</pattern>
```

The format attribute can optionally be specified and will control the parser used. Supported values are `"yaml"` (the default) and `"json"`.
    
The YAML parser can parse JSON but is slightly slower, thus specifying `format="json"` is recommended for JSON files.

Patterns having `<yaql>` tags will be considered non-matches if the `yaql` python module is not installed.

https://yaql.readthedocs.io/en/latest/getting_started.html


ogayot:
I'll need to gracefully handle the absence of the `yaml` module as well.

PR #486, opened 2025-03-27
test: add unshare crash test case

bdrung:
Add an unshare crash test case that is a stripped down version of a crash inside a container. This test case add code coverage to the `FileNotFoundError` exception code path in `forward_crash_to_container`. See also LP #2080499.

This merge request depends on https://github.com/canonical/apport/pull/484.


bdrung:
Running the unshare test case in a privileged Docker container works except one issue: The namespace is the same.

I'll leave it to the reader to figure out how to make it work in the container.

PR #489, opened 2025-04-16
test: move get_core_path tests to unit tests

bdrung:
The test case `test_get_core_path` fits into the unit test category. So move this test from the integration tests to the unit tests and split the test into individual tests (converting the comments into docstrings).


schopin-pro (changes requested):
The boot_id thing should be mocked out for those to be unit tests. I'd be fine keeping it in integration tests.

bdrung:
I mocked `get_boot_id()`.

PR #490, opened 2025-04-25
Include ELF package metadata from systemd-coredump

bdrung:
Include the ELF package metadata from systemd-coredump in the crash report. This information can be used later to check if the ELF package metadata matches the installed packages.

### DIFF
--- a/apport/report.py
+++ b/apport/report.py
@@ -12,6 +12,7 @@
 # pylint: disable=too-many-lines
 
 import atexit
+import collections
 import datetime
 import errno
 import fnmatch
@@ -19,6 +20,7 @@ import glob
 import grp
 import importlib.util
 import io
+import json
 import logging
 import os
 import pathlib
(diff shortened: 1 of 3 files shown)


apport/report.py
```diff
@@ -2019,6 +2021,14 @@ def _add_str_from_coredump(
         assert isinstance(value, str)
         self[key] = value.rstrip()
 
+    def _add_elf_package_metadata(self, coredump: dict[str, object]) -> None:
+        value = coredump.get("COREDUMP_PACKAGE_JSON")
+        if value is None:
+            return
+        assert isinstance(value, str)
+        decoded = json.loads(value, object_pairs_hook=collections.OrderedDict)
```
schopin-pro, line 2029: issue: after discussing it live it seems the round-trip is for pretty-printing + reproducibility, but that's really not clear from either code comments or commit log. Please fix that.

schopin-pro (changes requested):

apport/report.py
bdrung: Added a comment.

PR #491, opened 2025-05-09
ci: run the tests in verbose mode

schopin-pro:
This does not *completely* work around GHA's line buffering, but at least now when a test is hanging we have a better chance of knowing in which module that test is.


bdrung (approved):
The change looks good. The CI has red tests, but they are unrelated to this change.

PR #493, opened 2025-05-09
setup.py: remove the shebang mangling code

schopin-pro:
We don't need to do that anymore, Python2 is a thing of the past, everyone is on the python3 bandwagon.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,11 +35,13 @@ class clean_java_subdir(DistUtilsExtra.auto.clean_build_tree):
 
 
 # pylint: disable-next=invalid-name
-class install_fix_hashbangs(DistUtilsExtra.auto.install_auto):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -35,11 +35,12 @@ def run(self):
 
 
 # pylint: disable-next=invalid-name
-class install_fix_hashbangs(DistUtilsExtra.auto.install_auto):
-    """Fix hashbang lines in scripts in data dir."""
+class fix_symlinks_in_bash_completion(DistUtilsExtra.auto.install_auto):
```
bdrung, line 38: The name `fix_symlinks_in_bash_completion` does not reflect that it wraps the original install command.

bdrung (approved):

PR #495, opened 2025-05-09
report: let _build_proc_maps_cache return the cache

bdrung:
Let `_build_proc_maps_cache` return the proc maps cache. The caller in `_address_to_offset` can rely on it being not `None` then.


schopin-pro (commented):
suggestion: go a bit further and initialize `_proc_maps_cache` to an empty list to avoid the awkwardness of None-checking altogether.

PR #500, opened 2025-05-26
test: support coreutils rename to gnu-coreutils

bdrung:
coreutils 9.5-1ubuntu2 renames the binary package to gnu-coreutils and build with 'gnu' prefix. The source package coreutils-from is introduced to pick between gnu-coreutils and rust-coreutils. Several test cases fail due to this change.

Fix the test cases by one on three solutions:

* Drop test for coreutils (covered by the same test for bash)
* Replace binary by gzip
* Relax test to support gnu-coreutils and rust-coreutils


tests/integration/test_ui.py
```diff
@@ -820,8 +820,11 @@ def test_run_report_hanging(self) -> None:
             self.assertIn(expected_key, self.ui.report)
         self.assertEqual(self.ui.report["ProblemType"], "Hang")
         self.assertEqual(self.ui.report["ExecutablePath"], self.TEST_EXECUTABLE)
-        self.assertEqual(self.ui.report["Package"].split(" ")[0], "coreutils")
-        self.assertEqual(self.ui.report["SourcePackage"], "coreutils")
+        self.assertIn(
+            self.ui.report["Package"].split(" ")[0],
+            {"coreutils", "gnu-coreutils", "rust-coreutils"},
+        )
+        self.assertIn(self.ui.report["SourcePackage"], {"coreutils", "rust-coreutils"})
```
schopin-pro, line 827: question: Don't we have a third one based on busybox? IIRC it broke some builds recently?

bdrung: Good catch. I added busybox and toybox.

The tests fail on questing-proposed.

PR #503, opened 2025-05-27
apt_dpkg: check return type of pickle.load()

bdrung:
Check that `pickle.load()` returns a dictionary. Otherwise discard the loaded cache.

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -305,10 +305,11 @@ class _AptDpkgPackageInfo(PackageInfo):
             return self._virtual_mapping_obj
 
         mapping_file = os.path.join(configdir, "virtual_mapping.pickle")
-        if os.path.exists(mapping_file):
+        try:
             with open(mapping_file, "rb") as fp:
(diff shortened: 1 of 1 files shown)


apport/packaging_impl/apt_dpkg.py
```diff
@@ -308,9 +308,10 @@ def _virtual_mapping(self, configdir: str) -> dict[str, set[str]]:
         if os.path.exists(mapping_file):
             with open(mapping_file, "rb") as fp:
                 self._virtual_mapping_obj = pickle.load(fp)
-        else:
-            self._virtual_mapping_obj = {}
+            if isinstance(self._virtual_mapping_obj, dict):
+                return self._virtual_mapping_obj
```
schopin-pro, line 312: issue: this is stylistically *very* wrong. You shouldn't have your happy path as a deeply-indented early return!

schopin-pro (changes requested):

apport/packaging_impl/apt_dpkg.py
bdrung: Here is an alternative implementation. Do you prefer that?

PR #505, opened 2025-05-27
apt_dpkg: avoid double evaluation in get_source()

bdrung:
Query `self._apt_pkg` only once in `_AptDpkgPackageInfo.get_source()`.


bdrung:
IMO the failing codecov/patch can be ignored here.

PR #509, opened 2025-05-27
apt_dpkg: avoid double evaluation in get_architecture()

bdrung:
Query `self._apt_pkg` only once in
`_AptDpkgPackageInfo.get_architecture()`.


bdrung:
IMO we can ignore the failing codecov/patch.

PR #515, opened 2025-05-28
test: use cat and sleep from GNU coreutils

bdrung:
Several test cases use the sleep command as test case. Some kill the sleep command while it is sleeping and expect a certain stack trace. So enforce using the GNU coreutils implementation of the sleep command.

Since Ubuntu 25.10 (questing) the GNU implementation is shipped by gnu-coreutils and the commands are prefixed with `gnu`.

### DIFF
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -19,6 +19,14 @@ from unittest.mock import MagicMock
 import psutil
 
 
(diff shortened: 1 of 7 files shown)


bdrung:
Fixed test_install_packages_system and tested that locally.

PR #522, opened 2025-05-30
packaging: add type hints to all PackageInfo methods

bdrung:
Add type hints to all methods of `PackageInfo` and address mypy's complaints.


bdrung:
I suggest to ignore the codecov/patch missing the target percent.

PR #525, opened 2025-06-02
apport: avoid reading stat of /proc/<pid>/exe twice

bdrung:
Avoid reading the stat of `/proc/<pid>/exe` twice in `forward_crash_to_container`.

PR #528, opened 2025-06-02
apport: check the exe mtime within the proc root mount (LP: #2112272)

schopin-pro:
If the binary is in a container, our check will be looking at the binary OUTSIDE that container, which is rather problematic. We take advantage of the /proc/pid/root facility to avoid that.

Bug-Ubuntu: https://bugs.launchpad.net/ubuntu/+source/apport/+bug/2112272
Fixes: a806cac1fa13 ("apport: do consistency check before forwarding crashes")

### DIFF
--- a/data/apport
+++ b/data/apport
@@ -580,10 +580,11 @@ def forward_crash_to_container(
 
     # Validate that the crashed binary is owned
     # by the user namespace of the process
-    task_uid = proc_pid.stat("exe").st_uid
+    exe_path = f"root{proc_pid.readlink('exe')}"
+    exe_stat = proc_pid.stat(exe_path)
     try:
         with proc_pid.open("uid_map") as fd:
-            if not apport.fileutils.search_map(fd, task_uid):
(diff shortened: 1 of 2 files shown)


data/apport
```diff
@@ -914,7 +914,8 @@ def consistency_checks(
 
     # check if the executable was modified after the process started (e. g.
     # package got upgraded in between).
-    exe_mtime = os.stat("exe", dir_fd=proc_pid.fd).st_mtime
+    exe_path = os.readlink("exe", dir_fd=proc_pid.fd)
```
bdrung, line 917: You could use `proc_pid.readlink` here instead (and then `proc_pid.stat`).

bdrung:
Your merge requests contains https://github.com/canonical/apport/pull/525 now.

bdrung (approved):
I have successfully tested this change with a podman container:
* installed apport and chaos-marmosets in the container
* removed chaos-marmosets from the host
* run divide-by-zero in the container
* /var/log/apport.log on the host says: "crashed in a container without apport support" (thus reached the forwarding function)

PR #532, opened 2025-06-02
Import Report from apport.report directly

bdrung:
Import `Report` from `apport.report` directly as one step to lazy load the code in `apport/__init__.py` itself (and maybe deprecate it in the future).

### DIFF
--- a/apport/crashdb_impl/github.py
+++ b/apport/crashdb_impl/github.py
@@ -19,6 +19,7 @@ from dataclasses import dataclass
 from typing import Any, Self
 
 import apport.crashdb
+import apport.report
 
 
 class Github:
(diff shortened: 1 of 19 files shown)


schopin-pro:
I don't mind the PR itself, but your goal of emptying apport/__init__.py is worrying to me. You're basically planning to break all our downstream users, for... what reason?

And yes, we do have downstream users, and presumably not just the ones known by `reverse-depends`, as `cloud-init` doesn't seem to declare that relationship.

bdrung:
The goal is to lazy load the stuff in `apport/__init__.py`.

bdrung:
Here is the lazy loading PR:
* https://github.com/canonical/apport/pull/475
* https://github.com/canonical/apport/pull/533 with type hints to make mypy/pyright happy

PR #533, opened 2025-06-03
Lazy load elements in base apport module

bdrung:
For faster module loading and to avoid circular dependencies, load all elements in the base `apport` module in a lazy fashion.


bdrung:
Duplicate of https://github.com/canonical/apport/pull/475

PR #538, opened 2025-06-12
report: Handle FileNotFoundError when determining ExecutableTimestamp

bdrung:
There are several report of crashes:

```
  File "/usr/share/apport/apport", line 1304, in <module>
    sys.exit(main(sys.argv[1:]))
             ~~~~^^^^^^^^^^^^^^
  File "/usr/share/apport/apport", line 800, in main
    return process_crash_from_systemd_coredump(options.systemd_coredump_instance)
  File "/usr/share/apport/apport", line 1297, in process_crash_from_systemd_coredump
    report = apport.report.Report.from_systemd_coredump(coredump)
  File "/usr/lib/python3/dist-packages/apport/report.py", line 2085, in from_systemd_coredump
    report._add_executable_timestamp()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3/dist-packages/apport/report.py", line 813, in _add_executable_timestamp
    self["ExecutableTimestamp"] = str(int(os.stat(self["ExecutablePath"]).st_mtime))
                                          ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/.mount_crealiHIhwqX/bin/CrealityPrint'
```

The executable specified in the coredump might already be gone or not available (Apport might run in a service with `PrivateTmp=yes` set). Apport will discard crashes in `/tmp` anyway.

So catch `FileNotFoundError` and set `ExecutableTimestamp` to an error message.

Bug-Ubuntu: https://launchpad.net/bugs/2098792

### DIFF
--- a/apport/report.py
+++ b/apport/report.py
@@ -827,7 +827,13 @@ class Report(problem_report.ProblemReport):
             pass
 
     def _add_executable_timestamp(self) -> None:
-        self["ExecutableTimestamp"] = str(int(os.stat(self["ExecutablePath"]).st_mtime))
+        exe_path = self["ExecutablePath"]
+        try:
+            exe_stat = os.stat(exe_path)
+        except FileNotFoundError:
+            self["ExecutableTimestamp"] = f"Error: Executable '{exe_path}' not found"
+            return
+        self["ExecutableTimestamp"] = str(int(exe_stat.st_mtime))
 
     def add_proc_environ(
         self,
--- a/tests/unit/test_report.py
+++ b/tests/unit/test_report.py
@@ -1577,6 +1577,24 @@ No symbol table info available.
         self.assertEqual(report_output.getvalue().decode(), DIVIDE_BY_ZERO_REPORT)
         stat_mock.assert_called_once_with("/usr/bin/divide-by-zero")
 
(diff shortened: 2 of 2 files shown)


apport/report.py
```diff
@@ -827,7 +827,13 @@ def add_proc_info(
             pass
 
     def _add_executable_timestamp(self) -> None:
-        self["ExecutableTimestamp"] = str(int(os.stat(self["ExecutablePath"]).st_mtime))
+        exe_path = self["ExecutablePath"]
+        try:
+            exe_stat = os.stat(exe_path)
+        except FileNotFoundError:
+            self["ExecutableTimestamp"] = f"Error: Executable '{exe_path}' not found"
```
schopin-pro, line 834: issue: this field is used elsewhere, it'll just blow up there now.

schopin-pro (changes requested):

apport/report.py
bdrung: Good catch. I checked the code and found only one consumer. I changed that to handle this change now.

PR #543, opened 2025-06-20
ci: use dynamic container image names

bdrung:
Instead of using hard-coded codenames for the container images, use the tags that will update on their own. The current mapping is:

* debian:stable-slim -> bookworm
* ubuntu:latest -> noble
* ubuntu:rolling -> plucky
* ubuntu:devel -> questing

So this commit drops Ubuntu 24.10 "oracular" and adds Ubuntu 25.10 "questing".

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
     container:
-      image: ubuntu:plucky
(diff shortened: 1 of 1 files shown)


schopin-pro (changes requested):
NACK. Changing this manually isn't that costly, and we'd lose the property of having a somewhat stable and specific environment to run tests for a given commit.

PR #544, opened 2025-06-20
test: import apport-retrace to get coverage data

bdrung:
Calling `apport-retrace` as separate binary will prevent collecting code coverage data. So import the apport-retrace file and call the `main` function there.


laarmen (approved):
Arguably more importantly, it makes it much easier to debug when a test fails!

LGTM, assuming the tests end up passing? I don't have access to the test logs anymore.

bdrung:
The code coverage does not change.

PR #546, opened 2025-06-23
test: use true from GNU coreutils

bdrung:
Several test cases use the sleep command as test case. Some kill the sleep command while it is sleeping and expect a certain stack trace. So enforce using the GNU coreutils implementation of the sleep command.

Since Ubuntu 25.10 (questing) the GNU implementation is shipped by gnu-coreutils and the commands are prefixed with `gnu`.

Fixes: 6830e0cbd423 ("test: use cat and sleep from GNU coreutils")

### DIFF
--- a/tests/system/test_apport_valgrind.py
+++ b/tests/system/test_apport_valgrind.py
@@ -15,7 +15,7 @@ import subprocess
 import tempfile
 import unittest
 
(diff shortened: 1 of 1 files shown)


bdrung:
I tested this change on questing: https://github.com/bdrung/apport/actions/runs/15825014579/job/44602877376

PR #573, opened 2026-04-20
apport-cli: fix rounding printed upload percentage

bdrung:
Using `%u` on a float will truncate the decimals. So the percentage used to be rounded down to the next integer.

Fix the f-string to round the floating point down to an integer.

Fixes: 5a504336b2f6 ("refactor: Use f-strings everywhere")


Hyask (approved):
Sounds good to me :-)

PR #575, opened 2026-04-20
problem_report: fix file extension for zstd compressed values

bdrung:
When Apport is used with systemd-coredump, the CoreDump is zstd compressed (instead of gzip). When filing Launchpad bugs, the zstd compressed CoreDump is attached to the bug named `CoreDump.gz`.

Support zstd compressed values in `ProblemReport.write_mime`.

Tested with:

```
divide-by-zero
PYTHONPATH=$(pwd) APPORT_LAUNCHPAD_INSTANCE=qastaging bin/apport-cli /var/crash/_usr_bin_divide-by-zero.1000.crash
```

LP: #2148656

### DIFF
--- a/problem_report.py
+++ b/problem_report.py
@@ -102,12 +102,21 @@ def _base64_decoder(entry: Iterable[bytes]) -> Iterator[bytes]:
 
 
 def _create_compressed_attachment(name: str, value: bytes) -> email.mime.base.MIMEBase:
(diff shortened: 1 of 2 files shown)


Hyask (approved):
Thanks, that looks good to me.

bdrung:
Just rebased.

PR #577, opened 2026-04-20
Retrace proxy fix

seb128:


seb128:
Launchpad bug https://bugs.launchpad.net/apport/+bug/2148926

bdrung:
The linter is complaining. You can rebase your changes on top of https://github.com/canonical/apport/pull/578 to avoid running into `_fetch_contents_file` is too complex.

apport/packaging_impl/apt_dpkg.py
```diff
@@ -28,7 +28,6 @@
 import glob
 import gzip
 import hashlib
-import http.client
```
bdrung, line 31: There is another `http.client` usage in `apport/crashdb_impl` which might cause problems as well.

seb128: Things seem to be working without changing that file but maybe the retracer isn't hitting those codepath, do you have any idea how to to exercise this part of the code?

seb128:
Thanks, I will rebase once #578 lands

apport/packaging_impl/apt_dpkg.py
bdrung: This code is called by `upload_blob` which is called by `CrashDatabase.upload` which is called indirectly via `UserInterface.file_report`. So it looks like this code path is only hit by users and not by the retracer.

bdrung:
I rebased your changs on top of #578, added LP reference, adjusted the title, and pushed to https://github.com/bdrung/apport/commits/retrace-proxy-fix/

bdrung (approved):
Looks good and was already tested successfully.

PR #578, opened 2026-04-20
apt_dpkg: simplify _fetch_contents_file

bdrung:
Simplify `_fetch_contents_file` by removing the `update` variable and exit early in case there is no update needed.


Hyask (approved):
The unified diff is a bit confusing, but the split one made much more sense. Maybe it's just my brain being fried. Anyway, looks good, and welcome change, thanks!

bdrung:
The unified diff is harder to read because of the additional indentation change. So you are not alone.

bdrung:
Just rebased.

PR #580, opened 2026-04-21
problem_report: let load() return skipped keys

bdrung:
The documentation for `ProblemReport.load()` says:

> If binary is False, binary data is not loaded; the dictionary key is created, but its value will be an empty string.

Commit 64820a0e2de0 ("apport-unpack: Fix ValueError: ['separator'] has no binary content") changed this behavior to fix a crash in apport-unpack.

Setting non-loaded values to `None` will make type hints more complicated and less useful. Values in the problem report should not be allowed to be `None`.

Instead of adding dictionary keys with a special value for not loaded keys, just let `load()` return the keys that were skipped. This reduces guess work and make the code more robust.

Remove `ProblemReport.has_removed_fields()` since that information is lost and this function was only used in Apport tests.

### DIFF
--- a/bin/apport-unpack
+++ b/bin/apport-unpack
@@ -56,22 +56,17 @@ def open_report(report_filename: str) -> Iterator[(BinaryIO | gzip.GzipFile)]:
 
 def unpack_report_to_directory(
     report: problem_report.ProblemReport, target_directory: str
-) -> list[str]:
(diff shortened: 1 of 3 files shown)


Hyask (approved):

bdrung:
Just rebased.

PR #587, opened 2026-04-21
procutils: introduce parse_meminfo()

bdrung:
Reduce code duplication by introducing a `parse_meminfo()` function.

### DIFF
--- a/apport/procutils.py
+++ b/apport/procutils.py
@@ -1,0 +1,29 @@
+# Copyright (C) 2026 Canonical Ltd.
+# Author: Benjamin Drung <bdrung@ubuntu.com>
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation; either version 2 of the License, or (at your
+# option) any later version.  See http://www.gnu.org/copyleft/gpl.html for
+# the full text of the license.
+
+"""Functions to operate on files in /proc."""
+
+from collections.abc import Iterable
+
+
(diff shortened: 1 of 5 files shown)


apport/procutils.py
```diff
@@ -0,0 +1,29 @@
...
+# Free Software Foundation; either version 2 of the License, or (at your
+# option) any later version.  See http://www.gnu.org/copyleft/gpl.html for
+# the full text of the license.
+
+"""Functions to operate on files in /proc."""
+
+from collections.abc import Iterable
+
+
+def parse_meminfo(keys: Iterable[str]) -> dict[str, int]:
```
laarmen: suggestion: Don't filter the keys, and use a proper object rather than a dict as output.

Parsing this file is *really* cheap, so specifying the keys you want isn't terribly useful. OTOH, you already know which fields are important.

Sure, parsing those explicitly in the helper will make its implementation a bit uglier, but it will make the client code MUCH nicer to read:

```diff
-   meminfo = parse_meminfo({"Cached", "MemFree", "Writeback"})
-   return (meminfo["MemFree"] + meminfo["Cached"] - meminfo["Writeback"]) * 1024
+   meminfo = parse_meminfo()
+   return (meminfo.mem_free + meminfo.cached + meminfo.writeback) * 1024
```

laarmen (approved):
LGTM altogether, although could be even better ;)

apport/procutils.py
bdrung: Thanks for the review. I tried this suggestion and it looks nicer. So here you go with a the updated PR.

laarmen (approved):

PR #576, opened 2026-04-20
crashdb/launchpad: support CoreDump.zst

bdrung:
When Apport is used with systemd-coredump, the CoreDump is zstd compressed (instead of gzip).

Support retracing a core dump that is zstd compressed (and named `CoreDump.zst`).

Tested with:

```
divide-by-zero
PYTHONPATH=$(pwd) APPORT_LAUNCHPAD_INSTANCE=qastaging bin/apport-cli /var/crash/_usr_bin_divide-by-zero.1000.crash
PYTHONPATH=$(pwd) APPORT_LAUNCHPAD_INSTANCE=qastaging bin/apport-retrace --auth ~/.cache/apport/launchpad.credentials -S system 2143169
```

This PR is stacked on top of https://github.com/canonical/apport/pull/575

LP: #2148656

PR #590, opened 2026-04-22
problem_report: avoid double .txt file extension

bdrung:
Some bug reports like bug #2147517 have an attachment with double file extension: `CurrentDmesg.txt.txt`

Do not add `.txt` in case the key name already contains the file extension.

This PR is stacked on top of https://github.com/canonical/apport/pull/576

Bug: https://launchpad.net/bugs/2149892

### DIFF
--- a/tests/unit/test_problem_report.py
+++ b/tests/unit/test_problem_report.py
@@ -668,7 +668,7 @@ class T(unittest.TestCase):  # pylint: disable=too-many-public-methods
         pr["SimpleLineEnd"] = "bar\n"
         pr["TwoLine"] = "first\nsecond\n"
         pr["InlineMargin"] = "first\nsecond\nthird\nfourth\nfifth\n"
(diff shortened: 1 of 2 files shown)


tests/unit/test_problem_report.py
Hyask, line 760: I guess that one was throwing with the modification of the test without the modification of the code?

Hyask (approved):
Approved, assuming the answer to my question is somehow synonymous to "yes".

tests/unit/test_problem_report.py
bdrung: Yes, it will fail with:
```
>       self.assertEqual(parts[4].get_filename(), "Multiline.txt")
E       AssertionError: 'Multiline.txt.txt' != 'Multiline.txt'
E       - Multiline.txt.txt
E       ?              ----
E       + Multiline.txt
```

bdrung:
Just rebased.

PR #595, opened 2026-04-22
test: set locale for test_install_packages_error

bdrung:
The test `test_install_packages_error` will fail on systems where the locale is not set to C (english):

```
$ python3 -m pytest tests/integration/test_ui_gtk.py tests/system/test_packaging_apt_dpkg.py::test_install_packages_error
[...]
>       exc_info.match("E:Type 'bogus' is not known .* sources could not be read.$")
E       AssertionError: Regex pattern did not match.
E         Expected regex: "E:Type 'bogus' is not known .* sources could not be read.$"
E         Actual message: 'E:Typ »bogus« in Zeile 1 der Quellliste /tmp/tmp4erzi0jq/cache/Foonux 22.04/apt/etc/apt/sources.list ist unbekannt., E:Die Liste der Quellen konnte nicht gelesen werden.'
```

Set the locale explicitly for this test.

### DIFF
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -4,6 +4,7 @@ import contextlib
 import functools
 import importlib.machinery
 import importlib.util
+import locale
 import os
 import pathlib
 import shutil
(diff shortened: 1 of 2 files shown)


Hyask (approved):
Good idea!

PR #579, opened 2026-04-21
test: use pwd and ls from GNU coreutils in apport-valgrind tests

bdrung:
apport-valgrind fails on `ls` and `pwd` from uutils. The valgrind error output contains:

```
==868049== Invalid read of size 1
==868049==    at 0x532A166: strlen (vg_replace_strmem.c:506)
==868049==    by 0x41C323C: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x417E2C6: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x417E215: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x4318F5C: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x417E203: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x550E600: (below main) (libc_start_call_main.h:59)
==868049==  Address 0x7fff1a1eafcd is not stack'd, malloc'd or (recently) free'd
==868049==
==868049== Invalid read of size 1
==868049==    at 0x532A174: strlen (vg_replace_strmem.c:506)
==868049==    by 0x41C323C: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x417E2C6: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x417E215: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x4318F5C: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x417E203: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x550E600: (below main) (libc_start_call_main.h:59)
==868049==  Address 0x7fff1a1eafce is not stack'd, malloc'd or (recently) free'd
==868049==
==868049== Invalid read of size 1
==868049==    at 0x41C329D: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x417E2C6: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x417E215: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x4318F5C: ??? (in /usr/lib/cargo/bin/coreutils/ls)
==868049==    by 0x417E203: ??? (in /usr/lib/cargo/bin/coreutils/ls)
[...]

PR #598, opened 2026-04-27
ci: run unit/integration/system tests on ARM

bdrung:
Extend the CI to also run the unit/integration/system tests on arm64. Install gdb-multiarch on the system tests (needed on arm64 for retracing the amd64 crash).

To keep the amount of tests low, do not test `skip`, `unit-and-integration-installed`, and `system-installed` on ARM.

This PR depends on:
* https://github.com/canonical/apport/pull/579
* https://github.com/canonical/apport/pull/599
* https://github.com/canonical/apport/pull/600

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,10 +28,11 @@ env:
     bash binutils gcc gdb kmod libc6-dev libglib2.0-dev libxml2-utils
     polkitd python3-systemd valgrind xterm
   system_deps: >
-    dbus dbus-x11 gdb gvfs-daemons psmisc python3-dbus
+    dbus dbus-x11 gdb-multiarch gvfs-daemons psmisc python3-dbus
     ubuntu-dbgsym-keyring ubuntu-keyring xterm xvfb
   system_internet_deps: >
-    chaos-marmosets dpkg-dev gdb ubuntu-dbgsym-keyring ubuntu-keyring valgrind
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -70,14 +71,18 @@ jobs:
           - ubuntu:noble
           - ubuntu:questing
           - ubuntu:resolute
+        architecture:
+          - ubuntu-latest
+          - ubuntu-24.04-arm
```
Hyask, line 76: What's the strategy for picking `noble` here?
How can we also have something like `ubuntu-latest-arm`, and maybe try to get both the oldest supported series, and `devel`, to test on the widest possible range of software versions?

bdrung: noble is the latest supported (non-preview) version: https://github.com/actions/runner-images

There is no `ubuntu-latest-arm` or similar. Once there is a "latest" version, we should switch to it.

Since this upstream project targets the devel release, we should test on the latest version and not test on old releases.

We can test for old release on the SRU branches.

Hyask: Since this branch indeed targets mostly the devel release, is there a reason not to bump to 26.04 despite the preview? Given that this CI is not really relying on any pre-installed SDK in the images, I expect the "preview" flag to have very little effect here.

bdrung: No. I can do the resolute bump in a follow-up PR.

Hyask: Alright :-)

Hyask (approved):
Thanks!

PR #601, opened 2026-04-27
apt_dpkg: use str for contents mapping

bdrung:
Use `str` instead of `bytes` for the contents mapping. This speeds up the creation of the contents cache. On a Raspberry Pi 5 (running resolute) the creation time of a jammy contents mapping reduces by 16 % from 101 seconds to 85 seconds. The creation time on amd64 reduces by some percentage as well.


Hyask (approved):
Looks like a much welcome change. Both faster and nicer to maintain code.

PR #603, opened 2026-04-28
test: add unit tests for apport.report._read_open_fds()

bdrung:
Make the code coverage report stable by adding unit tests for `apport.report._read_open_fds()`.


Hyask (approved):
Thanks. Not sure why this makes the coverage more stable though. Was some other test non deterministic somehow? Like sometimes having `fdinfo` sometimes not?

bdrung:
I saw in other PRs that this exception code path in `_read_fd_entry` is only sometimes hit:

```python
     except OSError:
         return line
```

PR #612, opened 2026-04-28
apport: get value for CoreDump only once

bdrung:
Ease mypy to determine the type of the problem report value for `CoreDump` by getting the value only once and store it in a variable.


3v1n0 (approved):

bdrung:
Just rebased.

PR #620, opened 2026-04-30
report: decode crash output to string in add_kernel_crash_info

bdrung:
The functions `Report.crash_signature_addresses` and `Report._crash_signature_kernel_crash` expect the value for `Stacktrace` to be a string.

So let `add_kernel_crash_info` set the value for `Stacktrace` as string by decoding the `crash` output to UTF-8.

Bug: https://launchpad.net/bugs/2146806


Hyask (approved):
Thanks!

PR #622, opened 2026-04-30
apt_dpkg: arm64 is available on http://archive.ubuntu.com since questing

bdrung:
The arm64 package archive is available on archive.ubuntu.com since Ubuntu 25.10 "questing". So use this repository if available.

Bug: https://launchpad.net/bugs/2150685

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -187,14 +187,22 @@ def _load_all_sources(apt_dir: str) -> list[Deb822SourceEntry | SourceEntry]:
     return sources
 
 
-def _map_mirror_to_arch(uri: str, target_arch: str) -> str:
+def _availabe_on_archive_ubuntu_com(arch: str, dist: str) -> bool:
+    """Check if arch plus dist is available on http://archive.ubuntu.com."""
+    if arch == "arm64":
+        # arm64 is available on http://archive.ubuntu.com starting from questing on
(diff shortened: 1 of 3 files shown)


apport/packaging_impl/apt_dpkg.py
```diff
@@ -187,14 +187,22 @@ def _load_all_sources(apt_dir: str) -> list[Deb822SourceEntry | SourceEntry]:
     return sources
 
 
-def _map_mirror_to_arch(uri: str, target_arch: str) -> str:
+def _availabe_on_archive_ubuntu_com(arch: str, dist: str) -> bool:
+    """Check if arch plus dist is available on http://archive.ubuntu.com."""
+    if arch == "arm64":
+        # arm64 is available on http://archive.ubuntu.com starting from questing on
+        return dist not in {"trusty", "xenial", "bionic", "focal", "jammy", "noble"}
```
Hyask, line 194: `jammy` and `noble` are very very soon going to be copied to archive.u.c as well (possibly actually happening right now). Not blocking the approval on that though.

Hyask (approved):
I've left a comment and will let you decide the course of action. Wait? Anticipate? Leave it?

apport/packaging_impl/apt_dpkg.py
bdrung: Then let's wait for that since https://ports.ubuntu.com/ still provides arm64 for stonking. So there is no time pressure.

PR #623, opened 2026-05-04
report: Include source-and-location into Stacktrace and ThreadStacktrace

3v1n0:
Stacktraces in e.u.c and bugs are reported in a way that they provide lines for the specific package version, although sometimes it's not always clear what each call relates to.

This is because not all the errors are exposing the actual package version an error is related to, especially when similar traces are affecting very-different sources versions (and e.u.c does not care, since the processes still match).

This does not break the current parsing though and we added tests for ensuring this is the case.

Coming from: https://code.launchpad.net/~3v1n0/apport/+git/apport/+merge/504537

### DIFF
--- a/apport/report.py
+++ b/apport/report.py
@@ -1009,8 +1009,10 @@ class Report(problem_report.ProblemReport):
         ExecutablePath. This adds the following fields:
         - Registers: Output of gdb's 'info registers' command
         - Disassembly: Output of gdb's 'x/16i $pc' command
-        - Stacktrace: Output of gdb's 'bt full' command
-        - ThreadStacktrace: Output of gdb's 'thread apply all bt full' command
+        - Stacktrace: Output of gdb's 'bt -full -frame-info source-and-location'
+          command
+        - ThreadStacktrace: Output of gdb's
+          'thread apply all bt -full -frame-info source-and-location' command
         - StacktraceTop: simplified stacktrace (topmost 5 functions) for inline
           inclusion into bug reports and easier processing
         - AssertionMessage: Value of __abort_msg or __glib_assert_msg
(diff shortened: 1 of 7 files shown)


3v1n0:
We fail in 

[`apport/report.py:1035:89: E501 line too long (92 > 88 characters)`](https://github.com/canonical/apport/actions/runs/25329501026/job/74259195947?pr=623#step:5:25)

As the two linters seem to have different rules, and ruff does not really enforce a max line length, so what should we do there?

Hyask:
> As the two linters seem to have different rules, and ruff does not really enforce a max line length, so what should we do there?

If I were alone deciding that, I would leave only `ruff` enabled for code style, to avoid having situations like that where code that has been auto-formatted still hits formatting issues. Let's see what @bdrung thinks about it.

bdrung:
> As the two linters seem to have different rules

No, all linters are configured to have a 88 character line limit (ruff has this as default).

> and ruff does not really enforce a max line length, so what should we do there?

ruff is less strict about it and therefore does not complain. But the other tools are more strict. So please limit the line length to 88 characters.

bdrung:
Can you write the full URL instead of e.u.c in the git commit message?

tests/unit/test_report.py
```diff
@@ -881,6 +881,36 @@ def test_gen_stacktrace_top(self) -> None:
                 d (x=1) at crash.c:29"""),
         )
 
+        # frame-info source/location lines should not affect frame extraction
```
bdrung: Can you put that test into a separate unit test method?

apport/report.py
bdrung: Please update the documentation here as well.

3v1n0: Oh, I thought I did. Thanks for catching it

3v1n0:
> ruff is less strict about it and therefore does not complain. But the other tools are more strict. So please limit the line length to 88 characters.

Well it does since I got it reformatted...

3v1n0:
I feel that the ruff isn't reformatting these lines as per https://github.com/astral-sh/ruff/issues/8437 or similar, as basically:

```python3
        gdb_reports = {
            "Registers": "info registers",
            "Disassembly": "x/16i $pc",
            "Stacktrace": "bt -full -frame-info source-and-location",
            "ThreadStacktrace": "thread apply all bt -full -frame-info source-and-location",
            "AssertionMessage": "print __abort_msg->msg",
            "GLibAssertionMessage": "print (char*) __glib_assert_msg",
        }
```

The line of `ThreadStacktrace` does not get split automatically by the formatters.

bdrung:
The ruff CI jobs runs `ruff check` and not `ruff format`.

3v1n0:
I've added few more commits to this, as I think that to have proper retrace we should also ensure we have the sources, maybe this second part could be split out, but given it depends on the same base, I kept it in the same PR.

3v1n0:
ping?

PR #624, opened 2026-05-04
ui: avoid crashing when restart command is unavailable

JR-Morton:
## Summary

Fix the GTK Hang dialog so it does not offer "Relaunch this application"
when the report has no `ProcCmdline`.

The regular crash dialog already has this guard, but the Hang-specific
path did not. If the checkbox was selected, apport later called
`UserInterface.restart()`, which asserts that `ProcCmdline` exists.

## Notes

`Report.add_proc_info()` normally sets `ProcCmdline`, but reports can still
lack it if the field is removed by hooks or if the report was generated
outside the normal collection path. In that case the UI should not offer a
restart.

## Changes

- Only show the GTK Hang relaunch checkbox when `ProcCmdline` is present.
- Also require `ProcCmdline` before setting `response.restart`.
- Add a GTK system test covering a Hang report without `ProcCmdline`.
- Keep `UserInterface.restart()`'s assertion unchanged.

## Test

- `python3 -m unittest tests.unit.test_ui`
- `python3 -m unittest tests.system.test_ui_gtk`
- `git diff --check`

### DIFF
--- a/gtk/apport-gtk
+++ b/gtk/apport-gtk
@@ -229,8 +229,9 @@ class GTKUserInterface(apport.ui.UserInterface):
         if report_type == "Hang" and self.offer_restart:
             self.w("ignore_future_problems").set_active(False)
             self.w("ignore_future_problems").hide()
-            self.w("relaunch_app").set_active(True)
-            self.w("relaunch_app").show()
+            if self.can_restart():
+                self.w("relaunch_app").set_active(True)
+                self.w("relaunch_app").show()
             self.w("subtitle_label").show()
             self.w("subtitle_label").set_label(
                 "You can wait to see if it wakes up, or close or relaunch it."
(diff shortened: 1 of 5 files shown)


bdrung (commented):
Instead of making `restart` a no-op, could we change the UI instead to not allow restarting in case `ProcCmdline` is missing?

schopin-pro:
Subsidiary question: how did we get an empty field in the first place? Is
it a normal occurrence or an actual bug further up the data stream?

gtk/apport-gtk
```diff
@@ -229,8 +229,9 @@ class GTKUserInterface(apport.ui.UserInterface):
         if report_type == "Hang" and self.offer_restart:
             self.w("ignore_future_problems").set_active(False)
             self.w("ignore_future_problems").hide()
-            self.w("relaunch_app").set_active(True)
-            self.w("relaunch_app").show()
+            if "ProcCmdline" in self.report:
```
bdrung: Could we extent the UI object with an `can_restart` method that does this check?

bdrung (changes requested):
We need the same kind of change to kde/apport-kde.

bdrung:
I wasn't aware that we have `offer_restart`. This can make having `can_restart` and `offer_restart` confusing. The UI should only present the restart when we want to offer it plus apport can do that.

PR #628, opened 2026-05-19
ci: ensure bash is updated for integration tests

bdrung:
`get_package_origin("bash")` in `test_get_package_origin` might return `None` in case a version is installed that is not part of the archive any more.

Therefore ensure that `bash` is upgraded before running the integration tests. See also commit 9f68c5ff8e55 ("CI: Upgrade bash for integration tests").

Fixes: c6506e9856b9 ("ci: use environment variables for dependencies")


Hyask (approved):
LGTM

PR #629, opened 2026-05-19
Fix ProcMaps bytes/str crash from trailing NULs

rawrmonster17:
Title: Root Cause Analysis and proposed fix for bytes/ProcMaps null-termination

We traced the apport TypeError to ProcMaps values being left as bytes when '\x00' is present. This happens because `_try_unicode()` in apport 2.33.1 for Ubuntu does not force decoding for certain keys; upstream apport has `_STRING_KEYS` which avoids this. I prepared a patch that either centralizes null-stripping in `ProblemReport._try_unicode()` or adds defensive decoding+null-stripping in the four locations that parse ProcMaps. Unit tests pass locally and I've exercised the change against representative systemd-coredump outputs. Patch: lp2146806-bytes-procmaps.patch

Proposed action:
- Attach patch and request review from apport maintainers.
- If accepted upstream, request SRU to the Ubuntu apport packages and link the SRU to this bug.

Patch location: /tmp/lp2146806-bytes-procmaps.patch

### DIFF
--- a/apport/report.py
+++ b/apport/report.py
@@ -1855,7 +1855,10 @@ class Report(problem_report.ProblemReport):
         """
         if "ProcMaps" not in self or "Stacktrace" not in self or "Signal" not in self:
             return None
-        if self["ProcMaps"].startswith("Error: "):
+        proc_maps = self["ProcMaps"]
+        if isinstance(proc_maps, bytes):
+            proc_maps = proc_maps.decode("UTF-8", errors="replace")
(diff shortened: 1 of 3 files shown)


bdrung:
The mentioned `_STRING_KEYS` was added in commit 35881625520272b6f1ee947c0e6adbcec36b5f89 (links to https://launchpad.net/bugs/2146806). The plan is to backport this exact commit to Ubuntu 26.04 "resolute".

PR #630, opened 2026-06-23
ci: drop testing on Ubuntu 24.04 LTS "Noble Numbat"

bdrung:
Ubuntu 26.04 LTS "Resolute Raccoon" is the latest Ubuntu LTS version. Since the main branch targets the Ubuntu development release, drop testing on Ubuntu 24.04 LTS "Noble Numbat".


bdrung:
Finally found the reason for the failure: The Ubuntu 26.04 host uses amd64v3 which Apport fails to retrace.

PR #631, opened 2026-07-02
problem_report: submit JSON attachments as such

bdrung:
There might be fields like `pro-status.json` (from ubuntu-advantage-tools source hook) or `CoredumpPackageJson` that contain JSON data.

Upload those attachments with the correct file extension and with MIME type `application/json`.


Hyask (approved):
LGTM

bdrung:
Rebased and addressed ruff complaints: combined `name.endswith("Json") or name.endswith(".json")` to `name.endswith(("Json", ".json"))`.

PR #638, opened 2026-07-27
problem_report: use iterator in CompressedValue.write

bdrung:
Use the iterator version `decode_compressed_stream` in `CompressedValue.write` to avoid code duplication.

This is in preparation for https://github.com/canonical/apport/pull/639 which is not ready yet.


bdrung:
I am marking this PR as bug fix, because "problem_report: check if eof is reached in decode_compressed_stream" fixes silently failing.

Hyask (approved):
Alright, reading this commit by commit makes a lot more sense than the whole diff. LGTM.

PR #640, opened 2026-07-27
test: wait for child processes to complete execve()

bdrung:
`test_add_proc_info` fails with the parent cmdline content:

```
>       self.assertEqual(pr["ProcCmdline"], f"{catcmd} /foo\\ bar \\\\h \\\\\\ \\\\ -")
E       AssertionError: 'python3 -m pytest -ra -v -m not\\ require[37 chars]ion/' != '/usr/bin/gnucat /foo\\ bar \\\\h \\\\\\ \\\\ -'
E       - python3 -m pytest -ra -v -m not\ requires_internet tests/unit/ tests/integration/
E       + /usr/bin/gnucat /foo\ bar \\h \\\ \\ -
```

1. When `subprocess.Popen` spawns a process, it first calls `fork()` before calling `execve()`.
2. Immediately after `fork()`, the child process shares the parent's memory image.
3. During this brief window (before `execve()` completes), `/proc/{pid}/cmdline` contains the parent's command line.
4. The `wait_for_proc_exec` method checks if `/proc/{pid}/cmdline` has content. Because the parent's cmdline is non-empty, it return prematurely before `execve()` finishes replacing the process.
5. Apport then reads `/proc/{pid}/cmdline` while it still holds pytest's command line.

The same thing can happen in tests that use `run_test_executable`:

```
________________________ T.test_run_report_bug_pid_tags ________________________

self = <tests.integration.test_ui.T testMethod=test_run_report_bug_pid_tags>

    def test_run_report_bug_pid_tags(self) -> None:
        """run_report_bug() for a pid with extra tags"""
        with run_test_executable() as pid:
            # report a bug on text executable process
            argv = ["ui-test", "-f", "--tag", "foo", "-P", str(pid)]
            ui = UserInterfaceMock(argv)
            ui.present_details_response = apport.ui.Action(report=True)
            self.assertEqual(ui.run_argv(), True)

        assert ui.report
        self.assertIn("SourcePackage", ui.report)
        self.assertIn("Dependencies", ui.report)
        self.assertIn("ProcMaps", ui.report)
>       self.assertEqual(ui.report["ExecutablePath"], self.TEST_EXECUTABLE)
E       AssertionError: '/usr/lib/python3/dist-packages/pytest/__init__.py' != '/usr/bin/gnusleep'
E       - /usr/lib/python3/dist-packages/pytest/__init__.py
E       + /usr/bin/gnusleep

tests/integration/test_ui.py:795: AssertionError
```

Wait until the forked process changed the content of `cmdline`.

Bug: https://launchpad.net/bugs/2161888

### DIFF
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -170,6 +170,22 @@ def wait_for_sleeping_state(pid: int, timeout: float = WAITING_TIMEOUT) -> None:
     )
 
 
+def wait_for_proc_exec(pid: int, timeout_sec: float = WAITING_TIMEOUT) -> None:
+    """Wait for a spawned child process to populate /proc/<pid>/cmdline."""
+    elapsed_time = 0.0
+    while elapsed_time < timeout_sec:
+        with open(f"/proc/{pid}/cmdline", encoding="utf-8") as fd:
+            if fd.read():
+                return
+
+        time.sleep(0.1)
+        elapsed_time += 0.1
+
+    raise TimeoutError(
+        f"/proc/{pid}/cmdline not readable within {int(elapsed_time)} seconds."
+    )
+
+
 def wait_for_process_to_appear(
     process: str, already_running: Set[int], timeout: float = WAITING_TIMEOUT
 ) -> int:
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -101,6 +101,7 @@ def run_test_executable(
     if args is None:
         args = (get_gnu_coreutils_cmd("sleep"), "86400")
     with subprocess.Popen(args, env=env) as test_process:
+        wait_for_proc_exec(test_process.pid)
         try:
             yield test_process.pid
         finally:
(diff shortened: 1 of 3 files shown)


tests/helper.py
```diff
@@ -168,6 +169,31 @@ def wait_for_sleeping_state(pid: int, timeout: float = 5.0) -> None:
...
+
+    Immediately after fork(), /proc/{pid}/cmdline temporarily inherits the
+    parent test runner's command line. This method blocks until execve() has
+    replaced the process image and updated /proc/{pid}/cmdline to match the
+    child process's own command line.
+    """
+    parent_cmdline = pathlib.Path("/proc/self/cmdline").read_text(encoding="utf-8")
+    elapsed_time = 0.0
+    while elapsed_time < timeout_sec:
+        with open(f"/proc/{pid}/cmdline", encoding="utf-8") as fd:
```
Hyask, line 185: Why read with `Path.read_text` for `parent_cmdline` and with the explicit `open()` for `content`? Is `Path.read_text` somehow caching the content across loop iterations?

tests/helper.py
```diff
@@ -168,6 +169,31 @@ def wait_for_sleeping_state(pid: int, timeout: float = 5.0) -> None:
...
+            content = fd.read()
+            if content and content != parent_cmdline:
+                return
+
+        time.sleep(0.1)
+        elapsed_time += 0.1
+
+    raise TimeoutError(
+        f"/proc/{pid}/cmdline was not updated from parent cmdline"
+        f" within {int(elapsed_time)} seconds."
```
Hyask: nit: This "timeout" is actually not really time based, but rather counting a number of iterations. On a very slow system, that could very well become minutes. I don't really mind the method here, but would just suggest a less strict error message.
```suggestion
        f" within at least {int(elapsed_time)} seconds."
```

Hyask (changes requested):

tests/helper.py
bdrung: `parent_cmdline` contains the cmdline of the running process and is cached. The `open()` later opens the cmdline for the client process that we want to check. Should I rename `parent_cmdline` to `own_cmdline`?

tests/helper.py
bdrung: Good point. I'll change that. BTW, https://github.com/canonical/apport/pull/644 needs to be reviewed first to get back to a clean CI.

tests/helper.py
Hyask: No, what I meant is: why not use the `Path.read_text()` API everywhere?

bdrung: I kept the `open` call there to keep the unit tests simple. `open` is mocked, but `pathlib.Path` isn't. If I would use `pathlib.Path` in both cases, the mock would need to handle both.

bdrung:
Rebased and resolved rebase conflicts.

tests/helper.py
Hyask: Alright, fair enough.

Hyask (approved):

PR #643, opened 2026-07-28
rewrite check_files_md5 in pure Python

bdrung:
GNU coreutils 9.11 and rust-coreutils 0.9 changed `md5sum` to quote the filenames in case they contain spaces or other special characters. Example:

```
$ echo first > "with spaces"
$ echo second > "with'quotes"
$ echo third > 'with"quotes'
$ echo fourth > "with \" and ' quotes"
$ md5sum "with spaces" "with'quotes" 'with"quotes' "with \" and ' quotes" > sums
$ md5sum -c sums
'with spaces': OK
"with'quotes": OK
'with"quotes': OK
'with " and '\'' quotes': OK
```

This new quoting could be supported by checking for single or double quotes and use `shlex.split` in these cases.

To make the code more robust do not rely on the `md5sum` command and just use pure Python for it.

Bug: https://launchpad.net/bugs/2161957

### DIFF
--- a/apport/packaging_impl/apt_dpkg.py
+++ b/apport/packaging_impl/apt_dpkg.py
@@ -686,7 +686,7 @@ class _AptDpkgPackageInfo(PackageInfo):
             return []
 
         # create a list of files with a newer timestamp for md5sum'ing
-        sums = b""
+        modified = []
         sumfile = (
             f"/var/lib/dpkg/info/{package}:{self.get_system_architecture()}.md5sums"
         )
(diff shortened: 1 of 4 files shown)


apport/packaging_impl/apt_dpkg.py
Hyask: Not really related to your changes:
Wow, that logic is super brittle!
What if a filename contained a space? Like `/var/lib/dpkg/info/libolm3:amd64.md5sums` listing `usr/share/doc/libolm3/DH ratchet.svg` (very first example I could quickly find on my system).

apport/packaging_impl/apt_dpkg.py
```diff
@@ -711,17 +711,19 @@ def get_modified_files(self, package: str) -> list[str]:
                             "%s contains empty line, ignoring line", sumfile
                         )
                         continue
-                    s = os.stat(f"/{words[-1].decode('UTF-8')}".encode())
+                    expected_hash = words[0]
+                    filepath = pathlib.Path(f"/{words[-1]}")
```
Hyask: no blocking: Considering the example quoted above, do you want to take the occasion to change the `words[-1]` logic for something more reliable?

Hyask (approved):
Approved because your changes keep the current logic, but that function is super brittle and would need some rework.

apport/packaging_impl/apt_dpkg.py
bdrung: Valid point. I changed the logic to be more robust.

Hyask: I like your chosen solution, thanks :-)

Hyask (approved):
Great change!

bdrung:
Just rebased.

PR #644, opened 2026-07-29
Fix ruff 0.16.0 complaints

bdrung:
ruff 0.16.0 started to complain about much more than the previous version. Fix most of the complaints (one commit per class of complaint) and silence a few that are harder to address. Ignore them for now and let the future me fix those.

### DIFF
--- a/tests/unit/test_report.py
+++ b/tests/unit/test_report.py
@@ -214,10 +214,10 @@ ProcStatus:
 Signal: 4
 SignalName: SIGILL
 """
+# pylint: disable=line-too-long
 _DIVIDE_BY_ZERO_ENVIRONMENT = [
     "SHELL=/bin/bash",
-    "SESSION_MANAGER=local/hostname:@/tmp/.ICE-unix/6144"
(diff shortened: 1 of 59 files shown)


bdrung:
@schopin-pro If you want to review some Python code and learn what ruff is checking.

tests/unit/test_report.py
```diff
@@ -214,6 +214,7 @@
 Signal: 4
 SignalName: SIGILL
 """
+# ruff: ignore[ISC004] not better readable
```
Hyask: nit: For such things, I think I generally prefer simply having a long line. This brings way less confusion given the list becomes one item == one line.

Hyask (approved):

tests/unit/test_report.py
bdrung: Yes. That's the best readable solution. I changed this one commit.

bdrung: Can you review that "chore: wrap implicitly concatenated strings or unwrap" commit?

Hyask: Yes, that's much better imho.

PR #646, opened 2026-08-18
problem_report: validate key names in ProblemReport.load

bdrung:
Apport report files consist of key/value pairs based on the standard RFC822 format, except that Apport uses case sensitive keys. Key names must only contain numbers, letters, '.', '_', and '-'. This is validated in `ProblemReport.__setitem__`, but not when loading a report in `ProblemReport.load`.

So use `ProblemReport.__setitem__` in `ProblemReport.load` as well.

This PR depends on https://github.com/canonical/apport/pull/580 landing first.

Bug: https://launchpad.net/bugs/2161697

### DIFF
--- a/bin/apport-unpack
+++ b/bin/apport-unpack
@@ -91,14 +91,14 @@ def main() -> None:
             # the report needs to be loaded in one go.
             # The current implementation loads the whole report into memory.
             bin_keys = report.load(report_file, binary=args.report == "-")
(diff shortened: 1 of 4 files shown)


Hyask (approved):
Thanks for the quick patch!

bdrung:
I reworked the code a bit: I added `InvalidProblemReportKeyError` and a test case for loading a report with an invalid key.

Hyask (approved):
Looks clean that way :-)

PR #647, opened 2026-08-18
Fix partial writes for coredumps larger than 2 GiB

bdrung:
Python's `os.write()` is a low-level POSIX call that directly maps to the `write(2)` system call. The `write(2)` man page says:

> On Linux, write() (and similar system calls) will transfer at most 0x7ffff000 (2,147,479,552) bytes, returning the number of bytes actually transferred.

Because `os.write()` does not handle looping, writing raw `bytes` coredumps larger than 2 GiB results in truncated files of exactly 0x7ffff000 bytes, which are then reported as corrupted by gdb.

In `data/apport` loop over `os.write` until the whole coredump is written since that scripts operate on low level. In `apport.report` use `os.fdopen` to operate on a high-level file object to safely write all bytes.

Bug: https://launchpad.net/bugs/2109979

### DIFF
--- a/data/apport
+++ b/data/apport
@@ -263,6 +263,18 @@ def setup_signals():
     signal.signal(signal.SIGBUS, _log_signal_handler)
 
 
+def write_all(fd: int, data: bytes) -> None:
+    """Write all bytes to a file descriptor, handling potential partial writes."""
+    view = memoryview(data)
+    while view:
+        written = os.write(fd, view)
+        if written == 0:
+            raise OSError(f"os.write({fd}) returned 0 bytes written")
+        view = view[written:]
+
+
+# TODO: Split into smaller functions
+# pylint: disable-next=too-complex,too-many-locals
 def write_user_coredump(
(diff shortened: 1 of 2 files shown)


data/apport
```diff
@@ -328,11 +338,7 @@ def write_user_coredump(
                 os.close(core_file)
                 os.unlink(core_path, dir_fd=cwd)
                 return
-            if os.write(core_file, block) != size:
-                logger.error("aborting core dump writing, could not write")
-                os.close(core_file)
-                os.unlink(core_path, dir_fd=cwd)
-                return
+            write_all(core_file, block)
```
Hyask: This would now skip the closing and unlinking of the corefile if `write_all` throws. Isn't that a problem?

bdrung: Valid point. I changed both cases to catch OSError.

Hyask: Thanks!

Hyask (approved):

PR #649, opened 2026-08-28
apt_dpkg: keep :arch qualifier for foreign-arch packages (LP: #1784116)

varranvar:
The get_file_package() function unconditionally stripped the ":arch" qualifier from the matched dpkg .list filename. On multiarch systems, a file owned by a foreign-architecture package then resolved via the bare name to the native-arch package, which is often not installed, will cause a spurious package not installed error message (LP: #1784116).

This fix is to only strip the qualifier when it matches the system architecture and keep it for foreign architectures so the name resolves to the installed package.

### DIFF
--- a/tests/unit/test_packaging_apt_dpkg.py
+++ b/tests/unit/test_packaging_apt_dpkg.py
@@ -210,6 +210,47 @@ Components: main
             "/map_cachedir", impl.get_distro_codename(), "amd64"
         )
 
+    @unittest.mock.patch.object(impl, "_AptDpkgPackageInfo__fgrep_list_for_path")
+    @unittest.mock.patch.object(impl, "get_system_architecture")
+    @unittest.mock.patch("apport.packaging_impl.apt_dpkg.glob.glob")
+    @unittest.mock.patch("apport.packaging_impl.apt_dpkg.subprocess.run")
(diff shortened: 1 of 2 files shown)


tests/unit/test_packaging_apt_dpkg.py
```diff
@@ -210,6 +210,47 @@ def test_get_file_package_uninstalled_usrmerge(
...
+                "/lib/x86_64-linux-gnu/libc.so.6",
+                "libc6",
+            ),
+        ]
+
+        for list_file, queried_file, expected_package in test_data:
+            with self.subTest(list_file=list_file):
+                fgrep_list_for_path_mock.return_value = list_file
+                self.assertEqual(impl.get_file_package(queried_file), expected_package)
+
```
bdrung: Can you add a test to check that those mocks were called (to ensure that the mocking works)?

bdrung (commented):
Sorry for the long delay to review this change. It looks good to me except some smaller remarks.

* For consistency: Can you remove the LP reference from the title and add it to the commit footer?
```
Bug: https://launchpad.net/bugs/1784116
```
* Can you squash the test case into the commit that fixes the problem? IMO these belong together in an atomic commit?

PR #650, opened 2026-09-15
problem_report: validate key names in ProblemReport.extract_keys

bdrung:
Apport report files consist of key/value pairs based on the standard RFC822 format, except that Apport uses case sensitive keys. Key names must only contain numbers, letters, '.', '_', and '-'. This is validated in `ProblemReport.__setitem__`, but not when extracting a report in `ProblemReport.extract_keys`.

So use `_check_problem_report_key` in `ProblemReport.extract_keys` before contructing a path from the key.

Bug: https://launchpad.net/bugs/2167360


Hyask (approved):
Thanks

